PR #1, opened 2021-02-25
Add outline (section headings) for the patch subset portion of the specification.

garretrieger:
Also add definitions of primitive types and SparseBitSet.


svgeesus (approved):
All looks good to me

PR #2, opened 2021-03-22
Updated sparse bit set specification section.

garretrieger:
- Generalizes the structure so it can use branch factors other than the previously hardcoded 8. Using 4 is recommended for the general case, but I left open the option for encoders to chose factors other than 4.
- Added interval encoding using a 0 value node. My current thinking is that this does not add any additional size to the encoding even if it's not used if we recommend that the encoder should encode intervals inside of CompressedSet. So we should at least include it in the spec so that an advanced encoder could make use of it if desired.
- Adds an example that covers a branch factor of 4, with an encoded interval.

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -79,15 +79,15 @@ should be encoded by CBOR are given in the definition of those data types.
 ### SparseBitSet ### {#sparsebitset}
 
 A data structure which compactly stores a set of distinct unsigned integers. The set is represented as
-an tree where each node has a fixed number of children that recursively sub-divides an interval into
+a tree where each node has a fixed number of children that recursively sub-divides an interval into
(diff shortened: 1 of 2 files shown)


vlevantovsky (approved):
I think it would be a benefit to provide a link to external resource offering a general overview / illustration of a sparse bit set structure and bit arrays in general. I found a lot of online resources that discuss bit sets in code libraries but not a single one (other than one CS college course) that describes what they are. Would be nice to include a reference if we can find one.

garretrieger:
> I think it would be a benefit to provide a link to external resource offering a general overview / illustration of a sparse bit set structure and bit arrays in general. I found a lot of online resources that discuss bit sets in code libraries but not a single one (other than one CS college course) that describes what they are. Would be nice to include a reference if we can find one.

Good idea I'll see if I can find something.

PR #3, opened 2021-04-09
Fill out sections for New Font Request and Patch Font Request.

garretrieger:


Overview.bs
```diff
@@ -342,23 +343,133 @@ response_type can be one of the following values:
...
+*  [[#error-bad-request]]
+*  [[#error-internal-error]]
+
 Patch Font Request {#patch-request}
 -----------------------------------
 
+A patch font request is sent by a client to add data for additional codepoints to its font. Patch font
+requests are sent via HTTP and can only use the POST method. The request body is a single
+<a href="#PatchRequest"><code>PatchRequest</code></a> object encoded via CBOR. The fields of
+should be set as follows:
```
vlevantovsky, line 393: I suspect the reference to an object (PatchRequest?) are missing in this sentence.

Overview.bs
```diff
@@ -342,23 +343,133 @@ response_type can be one of the following values:
...
+*  <a href="#PatchRequest"><code>PatchRequest.indices_have</code></a>:<br/>
+    The set of codepoints that are covered by the client’s current copy of the font.
+    Specified using indices obtained by applying <code>codepoint ordering</code>
+    [[#codepoint-reordering]] to the set of codepoints.
+   
+*  <a href="#PatchRequest"><code>PatchRequest.indices_needed</code></a>:<br/>
+    The set of codepoints that the clients would like added to it’s copy of
+    the font. Specified using indices obtained by applying <code>codepoint ordering</code>
+    [[#codepoint-reordering]] to the set of codepoints.
+
```
vlevantovsky, line 423: I suggest that we should maintain a clear differentiation between _codepoints_ and _indices_, and not use these terms interchangeably. Indices are commonly associated with glyph IDs while codepoints are units of text content that, depending on a font used, can be associated with more than one glyph ID (contextual forms, stylistic sets, etc.).

vlevantovsky (approved):
Aside from my line-specific comments, one general topic we should keep in mind is the manifestation of requirements (i.e. using "must" vs. "MUST"). Capitalized MUST is what should be used to communicate a requirement of the spec, but there is a caveat - every MUST in the spec must be testable and _eventually_ covered by a corresponding conformance test (as opposed to using lowercase must as a unit of English language). This is why we should be mindful about the language we use when differentiating the spec requirements from common-sense mandates or external constraints.

garretrieger:
> Aside from my line-specific comments, one general topic we should keep in mind is the manifestation of requirements (i.e. using "must" vs. "MUST" according to https://tools.ietf.org/html/rfc2119). Capitalized MUST is what should be used to communicate a requirement of the spec, but there is a caveat - every MUST in the spec must be testable and _eventually_ covered by a corresponding conformance test (as opposed to using lowercase must as a unit of English language). This is why we should be mindful about the language we use when differentiating the spec requirements from common-sense mandates or external constraints.

As discussed we'll leave these lower case throughout. I checked over all usages of must in the spec so far and removed a few that weren't really needed.

Overview.bs
garretrieger: I reworded this slightly to try and make it more clear.

PR #4, opened 2021-05-14
Simplifies the client specification and adds server specification for patch subset

garretrieger:
- Remove the response_type field from responses. This is now inferred by the client based on which fields are set/unset.
- Unify handling of patch and rebase in the client (treats rebase as a patch against a zero length file).
- Adds specification for expected server behaviour. This is intentionally very minimal to give server implementations plenty of flexibility.

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -449,6 +449,8 @@ should interpret and process the fields of the object as follows:
     resend the request that triggered this response but use the new codepoint ordering provided in
     this response.
 
+TODO(garretrieger): how to handle error status codes (ie. 4xx, 5xx).
+
 ### Client Side Checksum Mismatch ### {#client-side-checksum-mismatch}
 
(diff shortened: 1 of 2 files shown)


vlevantovsky (approved):
Overall, the changes look good. I think that once we get to a more stable version of the spec, we should have a group discussion reviewing all "musts" to figure out which one can be testable and, therefore, capitalized. Certain "musts" (like e.g. protocol version must be set to 0, patch format must contain at least one format, etc.) can be easily tested by malformed inputs.

garretrieger:
> Overall, the changes look good. I think that once we get to a more stable version of the spec, we should have a group discussion reviewing all "musts" to figure out which one can be testable and, therefore, capitalized. Certain "musts" (like e.g. protocol version must be set to 0, patch format must contain at least one format, etc.) can be easily tested by malformed inputs.

Sounds good to me.

PR #11, opened 2021-07-21
Add the capability for a client to incrementally transfer vf axis space

garretrieger:
Similar to codepoints this adds an axis space have and axis space needed to the request. This allows a client to get a subset of the font with a reduced axis space and then later on expand the axis space via a patch.

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -385,7 +385,7 @@ For an <code>AxisInterval</code> object to be well formed:
 
 *  <code>start</code> must be set.
 
-*  <code>end</code> optional, must be greater than <code>start</code>. If <code>end</code> is not set
+*  <code>end</code> is optional, must be greater than <code>start</code>. If <code>end</code> is not set
     then this interval is a single point, <code>start</code>.
 
 #### PatchRequest #### {#PatchRequest}
--- a/Overview.bs
+++ b/Overview.bs
@@ -9,8 +9,8 @@ ED: https://w3c.github.io/PFE/Overview.html
 Editor: Chris Lilley, W3C, https://svgees.us/, w3cid 1438
 Editor: Myles C. Maxfield, Apple Inc., mmaxfield@apple.com, w3cid 77180
 Editor: Garret Rieger, Google Inc., grieger@google.com
-Abstract: Example example
-Status Text: This is a largely empty document because we have just started working on it.
+Abstract: TODO(garretrieger)
+Status Text: initial draft is in progress.
 </pre>
 
 <!--
@@ -42,6 +42,19 @@ Status Text: This is a largely empty document because we have just started worki
         "title": "Shared Brotli Compressed Data Format",
         "date": "27 Jul 2021"
     }
+	}
+</pre>
+
+<pre class=biblio>
+	{
+	    "OpenType-Variations": {
+            "href": "https://docs.microsoft.com/en-us/typography/opentype/spec/otvaroverview",
+            "authors": [],
+            "status": "Note",
+            "publisher": "Microsoft",
+			"title": "OpenType Font Variations Overview",
+			"date": "23 October 2020"
+		}
 	}
 </pre>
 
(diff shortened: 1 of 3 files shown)


Overview.bs
```diff
@@ -333,6 +357,24 @@ The list of ranges is encoded as a series of deltas. For example the ranges
...
+
+<code>AxisInterval</code> defines an interval (from <code>start</code> to <code>end</code>) on some
+variable axis in a font.
+
+For an <code>AxisInterval</code> object to be well formed:
+
+*  <code>start</code> must be set.
+
+*  <code>end</code> optional, must be greater than <code>start</code>. If <code>end</code> is not set
+    then this interval is a single point, <code>start</code>.
```
vlevantovsky, line 376: Minor correction: should probably be "`end` is optional"

Overview.bs
```diff
@@ -601,7 +700,7 @@ the new value for that codepoint.
 A server is free to choose any codepoint ordering, but should try to pick one that will minimize the
```
vlevantovsky, line 700: General comment: one thing I noticed is that the current version of the spec seem to be written as a document that describes a particular PFE implementation. The alternative way is to describe what implementations [must / must not / should / should not] do, and all these requirement need to be clearly identified as normative (and, therefore, testable).
If something is expected to be done a particular way (e.g. checksum calculation below), it needs to be defined as normative - the difference is whether something can be done this way or MUST be done this way.

vlevantovsky (changes requested):
The changes related to VF axis space are fine, but overall - the more language is added, the harder it gets to make the text of the spec normative. There are quite a few "must" used as part of the descriptive text, and some of them are likely testable and need to be made normative (i.e., we need to use "MUST" instead, e.g. when we specify what it means for an object to be "well-formed").

svgeesus (approved):
This all looks good to me.

svgeesus:
> 
> 
> The changes related to VF axis space are fine, but overall - the more language is added, the harder it gets to make the text of the spec normative. There are quite a few "must" used as part of the descriptive text, and some of them are likely testable and need to be made normative (i.e., we need to use "MUST" instead, e.g. when we specify what it means for an object to be "well-formed").

Yes, good point. In the WOFF 1 and 2 specifications we marked up individual testable statements, so that they could be linked to from tests (and so that we could do test coverage analysis).

That isn't needed for First Public Working Draft but will be needed down the line.

vlevantovsky:
> That isn't needed for First Public Working Draft but will be needed down the line.

Agree. My only concern was that the more language is added, the more work it will be going back to revise it. By capitalizing "musts" that represent testable requirements we will at least reduce the time needed to update the spec later.

garretrieger:
> > That isn't needed for First Public Working Draft but will be needed down the line.
> 
> Agree. My only concern was that the more language is added, the more work it will be going back to revise it. By capitalizing "musts" that represent testable requirements we will at least reduce the time needed to update the spec later.

In the "Document Conventions" section we state: "However, for readability, these words do not appear in all uppercase letters in this specification."  so I've been intentionally not capitalizing the should/must/should nots/must nots, but wherever I use must/should I've generally done so intentionally with the meaning from the definitions here: https://datatracker.ietf.org/doc/html/rfc2119

garretrieger:
> > > That isn't needed for First Public Working Draft but will be needed down the line.
> > 
> > 
> > Agree. My only concern was that the more language is added, the more work it will be going back to revise it. By capitalizing "musts" that represent testable requirements we will at least reduce the time needed to update the spec later.
> 
> In the "Document Conventions" section we state: "However, for readability, these words do not appear in all uppercase letters in this specification." so I've been intentionally not capitalizing the should/must/should nots/must nots, but wherever I use must/should I've generally done so intentionally with the meaning from the definitions here: https://datatracker.ietf.org/doc/html/rfc2119

If we're going to use capitalized keywords then let's remove that sentence from the document conventions as well.

vlevantovsky:
> In the "Document Conventions" section we state: "However, for readability, these words do not appear in all uppercase letters in this specification." so I've been intentionally not capitalizing the should/must/should nots/must nots, but wherever I use must/should I've generally done so intentionally with the meaning from the definitions here: https://datatracker.ietf.org/doc/html/rfc2119

Ah, thank you, I completely missed this point. I was not aware the established process would allow us to essentially redefine RFC 2119 keywords from capitalized to lowercase, and didn't pay much attention to formal parts of the spec.

> If we're going to use capitalized keywords then let's remove that sentence from the document conventions as well.

I guess this is a question for Chris - can we redefine the RFC2119 keywords to have the same meaning when they are used in lowercase? If yes, then everything will stay the same and no further changes are needed. If however, the keywords must be capitalized (my expectation was that this is the case), then we would need to make changes at some point of document development.

vlevantovsky (approved):

PR #16, opened 2021-08-13
Collection of cleanups to recently added compressed int list sections

garretrieger:
- Renames CompressedIntList to IntegerList and SortedIntegerList
- Fix up some of the examples which had incorrect example bytes.
- Use consistent capitalization in type names.
- Some re-ordering of examples and text.


vlevantovsky (approved):
LGTM

PR #17, opened 2021-08-20
A few misc fixes

garretrieger:
- Last UIntBase128 example was incorrect. Fix it.
- Add example checksums.
- Explicitly state the integer bounds for members of IntegerList, SortedIntegerList, and RangeList.


vlevantovsky (commented):
These changes themselves look fine, but now (paying more attention to data types in general) I am having difficulties trying to reconcile these lists with the declared much larger primitive integer data type being defined as "An integer value range [-2<sup>64</sup> - 1, 2<sup>64</sup> - 1] inclusive." The integer range itself is suspect (should it at least be 2<sup>63</sup> for signed int64?)

garretrieger:
> These changes themselves look fine, but now (paying more attention to data types in general) I am having difficulties trying to reconcile these lists with the declared much larger primitive integer data type being defined as "An integer value range [-264 - 1, 264 - 1] inclusive." The integer range itself is suspect (should it at least be 263 for signed int64?)

The larger range is the range of the native cbor integer type. The 2^64 bounds comes about because CBOR has a distinct type for positive integers and a distinct type for negative integers. Thus the positive and negative range are each encoded into a full uint64.

vlevantovsky (approved):

PR #18, opened 2021-08-20
Update AxisSpace to use a list of intervals for each tag.

garretrieger:
This allows for a client to specify things like I want weight 300 and weight 700 without being forced to ask for the entire interval from 300 to 700.


vlevantovsky (approved):
Looks good, thanks.

Issue #31, opened 2021-09-08
Privacy section is entirely missing


litherum:
https://w3c.github.io/IFT/Overview.html#priv

> No Privacy issues have been raised against this document

This statement is false. The document needs to describe how clients must request additional unnecessary glyphs in order to dilute the signal of what content the user is viewing. Ideally, we would place some minimum requirements here to make sure a browser can't just request 0 unnecessary glyphs and claim conformance.

garretrieger:
Right, I'll make a pass at adding that to the privacy section.

svgeesus:
Okay [here we go](https://w3c.github.io/IFT/Overview.html#priv)

PR #47, opened 2021-11-18
First draft of range request method

litherum:
First draft of range request method

### DIFF
--- a/RangeRequest.bs
+++ b/RangeRequest.bs
@@ -22,7 +22,7 @@ Incremental Font Transfer (IFT) is a collection of technologies to improve the l
 
 There are two technologies inside the IFT collection of technologies. This document describes one of them: the "Range-request method." In this method, the browser uses HTTP range requests to download only specific regions of the font file that it needs. In order for this method to be as effective as possible, the font file itself should be arranged in a particular way, in order to decrease the number of requests the browser needs to make.
 
-The range-request IFT method was specifically designed to target good performance on Chinese, Japanese, and Korean languages. These languages have the largest font files, and include many glyphs which are generally independent. Thus, these languages are a good target for selective downloads.
+The range-request IFT method has best performance on Chinese, Japanese, and Korean languages. These languages usually have the largest font files, and usually include many glyphs which are generally independent.
 
 The range-request method was designed to be as easy as possible to deploy, and doesn't require any server-side software. Deploying this method involves two steps:
 
@@ -34,6 +34,8 @@ A rearranged font file is still a valid font file following existing formats; th
 <h2 id="opt-in">
 Opt-In Mechanism</h2>
 
+<em>This section is general to both IFT methods, and is not specific to the range-request method.</em>
+
 The collection of IFT technologies utilizes a single shared opt-in mechanism. Each method does not use its own opt-in mechanism; instead, the webpage opts into the IFT technologies as a whole, and the browser and server negotiate to decide which specific method will be employed.
 
 The opt-in mechanism is the <code>incremental</code> keyword inside the ''@font-face'' block. Websites specify this keyword inside their <code>@font-face</code> block, and the browser is then responsible for using IFT technologies, and for negotiating with the server to determine which specific IFT method to use.
@@ -43,7 +45,7 @@ The use of the <code>incremental</code> keyword in this CSS rule indicates to th
 <pre>
 @font-face {
     font-family: "MyCoolWebFont";
-    src: url("MyCoolWebFont.otf") format(opentype supports incremental);
+    src: url("MyCoolWebFont.otf") tech(incremental);
 }
 </pre>
 </div>
@@ -53,10 +55,12 @@ Note: Each individual <code>@font-face</code> block may or may not opt-in to IFT
 <h2 id="method-selection">
 IFT Method Selection</h2>
 
-When a page has indicated that a particular font should utilize IFT technology, the browser must determine which method to use. Different browsers may support different IFT methods, and different servers may support different IFT methods, so a negotation occurs at such:
+<em>This section is general to both IFT methods, and is not specific to the range-request method.</em>
+
+When a page has indicated that a particular font should utilize IFT technology, the browser must determine which method to use. Different browsers may support different IFT methods, and different servers may support different IFT methods, so a negotation occurs as such:
 
 1. The browser makes the first request to the server. If the client prefers the <a href="https://w3c.github.io/IFT/Overview.html#patch-incxfer">patch-subset method</a>, it sends the relevant <a href="https://w3c.github.io/IFT/Overview.html#negotiating-transfer-type">query parameter</a>. If the client prefers the range-request method, it does not send the query parameter.
-2. If the server receives the patch-subset query parameter and wishes to honor it, the server may reply with the <a href="https://w3c.github.io/IFT/Overview.html#handling-patch-response">patch-subset magic number</a>. Otherwise, the server may reply with the [[!RFC7233]] <a href="https://httpwg.org/specs/rfc7233.html#header.accept-ranges"><code>Accept-Ranges</code></a> header.
+2. If the server receives the patch-subset query parameter and wishes to honor it, the server may reply with a valid <a href="https://w3c.github.io/IFT/Overview.html#PatchResponse">patch subset response</a> which includes the <a href="https://w3c.github.io/IFT/Overview.html#handling-patch-response">patch-subset magic number</a>. Otherwise, the server may reply with the [[!RFC7233]] <a href="https://httpwg.org/specs/rfc7233.html#header.accept-ranges"><code>Accept-Ranges</code></a> header.
 3. If the client receives the patch-subset magic number, it commences using the patch-subset method. Otherwise, if the client receives the <code>Accept-Ranges: bytes</code> header, it commences using the range-request method. Otherwise, the whole font file is downloaded, and the current non-incremental loading behavior is used.
 
 <h3 id="fallback">
@@ -75,6 +79,9 @@ This summarizes behaviors that result from the above method selection.
     <tbody>
         <tr><th>Server supports both range-request method and patch-subset method
             <td>Client makes initial request without query parameter, and possibly with the <code>Range</code> header. Because all patch-subset servers must support the range-request method, the server replies with <code>Accept-Ranges</code> and initial font data. Client/server commence using range-request method.
+
+            Issue: <a href="https://github.com/w3c/IFT/issues/56">Server-recommended IFT Method selection</a>
+
             <td>Client makes initial request with query parameter. Server replies with the patch-subset magic number, and client/server commence using patch-subset method.
         <tr><th>Server supports only range-request method
             <td>Same as above.
@@ -94,6 +101,8 @@ Background</h2>
 
 A particular organization of font files is beneficial for improving the performance of the range-request IFT method. The range-request IFT method only works with [[!TRUETYPE]], [[!OPENTYPE]], [[!WOFF]], or [[!WOFF2]] files. All of these file formats use an <a href="https://docs.microsoft.com/en-us/typography/opentype/spec/otff#organization-of-an-opentype-font"><code>sfnt</code> wrapper</a> which provides a directory of tables inside the font file. A <code>sfnt</code>-based font file is mainly composed of a collection of independent tables.
 
+Issue: <a href="https://github.com/w3c/IFT/issues/59">Using WOFF2 files with range-request mechanism doesn't seem to be a viable option</a>
+
 <h3 id="font-organization-introduction">
 Introduction</h2>
 
@@ -102,7 +111,7 @@ The term <dfn>range-request optimized font</dfn> is used to describe a font file
 Note: Because optimizing a font does not change its file format, no new additional tooling is necessary to interact with these optimized fonts. They are still valid fonts, but with a particular internal organization.
 
 <div class=example>
-The result of optimizing an OpenType font for the range-request IFT method is still a valid OpenType font.
+The result of optimizing an OpenType font for the range-request IFT method is still a valid OpenType font. The resulting file may be larger (by byte count) than it was before optimizing it, but fewer of those bytes should be necessary for a client to download in order to render a target text.
 </div>
 
 Note: There are no <code>MUST</code>-level requirements on the organization of a [=range-request optimized font=]. Any arbitrary font file may be considered to be a [=range-request optimized font=]. However, additional optimizations should increase the performance of loading in a browser via the range-request IFT method. Font creators are encouraged to enact as many of the optimizations listed in this section as are reasonable for the fonts they create.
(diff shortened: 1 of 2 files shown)


RangeRequest.bs
```diff
@@ -10,26 +10,218 @@ Editor: Chris Lilley, W3C, https://svgees.us/, w3cid 1438
...
 </pre>
 
+<h2 id="intro">
+Introduction</h2>
 
-Introduction {#intro}
-=====================
+<em>This section is not normative.</em>
+
+Incremental Font Transfer (IFT) is a collection of technologies to improve the latency of remote fonts (or "web fonts") on the web. Without this technology, a browser needs to download every last byte of a font before it can render any characters using that font. IFT allows the browser to download only some of the bytes in the file, thereby decreasing the perceived latency between the time when a browser realizes it needs a font and when the necessary text can be rendered with that font.
```
svgeesus, line 21: Meta: see [Path Forwards: one, two, three specs?](https://github.com/w3c/IFT/issues/48)

RangeRequest.bs
```diff
@@ -10,26 +10,218 @@ Editor: Chris Lilley, W3C, https://svgees.us/, w3cid 1438
...
+
+There are two technologies inside the IFT collection of technologies. This document describes one of them: the "Range-request method." In this method, the browser uses HTTP range requests to download only specific regions of the font file that it needs. In order for this method to be as effective as possible, the font file itself should be arranged in a particular way, in order to decrease the number of requests the browser needs to make.
+
+The range-request IFT method was specifically designed to target good performance on Chinese, Japanese, and Korean languages. These languages have the largest font files, and include many glyphs which are generally independent. Thus, these languages are a good target for selective downloads.
```
svgeesus, line 25: This doesn't seem correct, because for slower networks the performance is worse.

RangeRequest.bs
```diff
@@ -10,26 +10,218 @@ Editor: Chris Lilley, W3C, https://svgees.us/, w3cid 1438
...
+Issue: https://github.com/w3c/IFT/issues/28 What about font collections?
+
+<h3 id="font-organization-glyph-independence">
+Glyph Independence</h2>
+
+Note: The goal of this section is to make every glyph independent from each other.
+
+A [=range-request optimized font=] should not use <a href="https://developer.apple.com/fonts/TrueType-Reference-Manual/RM03/Chap3.html">TrueType instructions</a>.
+
+Note: Some rasterizer software does not even execute these instructions in the first place.
```
svgeesus, line 148: But some does and looks awful without it.
What is the rationale for this constraint?

RangeRequest.bs
```diff
@@ -10,26 +10,218 @@ Editor: Chris Lilley, W3C, https://svgees.us/, w3cid 1438
...
-==============================
+Note: Another valid alternative is to treat the entire font as residing on an asynchronous virtual filesystem, and have the browser track which ranges of the font it ended up reading during its normal operation. The browser could then request those regions in range requests.
 
-More to come soon...
+<h2 id="sec">
+Security Considerations</h2>
 
-<h2 class=no-num id=priv-sec>Privacy and Security Considerations</h2>
+<h2 id="priv">
+Privacy Considerations</h2>
```
svgeesus, line 220: This should not be empty. See https://github.com/w3c/IFT/issues/31

RangeRequest.html
```diff
@@ -4,10 +4,9 @@
   <meta content="width=device-width, initial-scale=1, shrink-to-fit=no" name="viewport">
```
svgeesus, line 4: (review based on .bs while reading the html; no review comments on generated file)

svgeesus (commented):
Review left as "Comment" rather than "Approve" pending discussion of https://github.com/w3c/IFT/issues/48

However I am in favor of approving, and fixing afterwards.

RangeRequest.bs
vlevantovsky: Also, the claim that CJK fonts "include many glyphs which are generally independent" may not hold for fonts that have large number of composite glyphs.

RangeRequest.bs
```diff
@@ -10,26 +10,218 @@ Editor: Chris Lilley, W3C, https://svgees.us/, w3cid 1438
...
+
+<table class="data">
+    <thead>
+        <tr>
+            <th>&nbsp;
+            <th>Client prefers range-request method
+            <th>Client prefers patch-subset method
+    <tbody>
+        <tr><th>Server supports both range-request method and patch-subset method
+            <td>Client makes initial request without query parameter, and possibly with the <code>Range</code> header. Because all patch-subset servers must support the range-request method, the server replies with <code>Accept-Ranges</code> and initial font data. Client/server commence using range-request method.
```
vlevantovsky, line 77: Considering the data we collected during PFE evaluation work on how network conditions affect the font transfer efficiency, and assuming that servers supporting both methods would eventually learn how to optimize data transfer based on those conditions - should we allow a server to _suggest_ a recommended approach? E.g. when a client wants to opt in for incremental transfer but the server recommends full font load, or when a client initially opts in for "range-request" but the server recommends "patch-subset" (or vice versa).

RangeRequest.bs
```diff
@@ -10,26 +10,218 @@ Editor: Chris Lilley, W3C, https://svgees.us/, w3cid 1438
...
+<h2 id="font-organization">
+Font organization</h2>
+
+<h3 id="font-organization-background">
+Background</h2>
+
+<em>This section is not normative.</em>
+
+A particular organization of font files is beneficial for improving the performance of the range-request IFT method. The range-request IFT method only works with [[!TRUETYPE]], [[!OPENTYPE]], [[!WOFF]], or [[!WOFF2]] files. All of these file formats use an <a href="https://docs.microsoft.com/en-us/typography/opentype/spec/otff#organization-of-an-opentype-font"><code>sfnt</code> wrapper</a> which provides a directory of tables inside the font file. A <code>sfnt</code>-based font file is mainly composed of a collection of independent tables.
+
```
vlevantovsky, line 96: I get it how range-request can be applied to TTF, OTF and WOFF (on a per-table basis), but I'm not 100% sure I understand how range-request method works with WOFF2 file (that is composed of multiple data streams). This section may need to be extended to include additional details/clarifications.

RangeRequest.bs
```diff
@@ -10,26 +10,218 @@ Editor: Chris Lilley, W3C, https://svgees.us/, w3cid 1438
...
+Note: The goal of this section is to make every glyph independent from each other.
+
+A [=range-request optimized font=] should not use <a href="https://developer.apple.com/fonts/TrueType-Reference-Manual/RM03/Chap3.html">TrueType instructions</a>.
+
+Note: Some rasterizer software does not even execute these instructions in the first place.
+
+A [=range-request optimized font=] should not use <a href="https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6glyf.html">Compound glyphs</a>.
+
+Note: Compound glyphs can be flattened by inlining their component glyphs to become additional contours.
+
```
vlevantovsky, line 153: Fonts that were efficiently designed using composite glyphs _can be flattened_ but that can easily double the overall font file size. It seems counterproductive to require that we double the original font size only to make it eligible for a particular incremental font transfer method that, in the end, may not even offer a tangible benefit - a client may be better off downloading the whole original font from the get go.

RangeRequest.bs
```diff
@@ -10,26 +10,218 @@ Editor: Chris Lilley, W3C, https://svgees.us/, w3cid 1438
...
+A [=range-request optimized font=] should not use <a href="https://wwwimages2.adobe.com/content/dam/acom/en/devnet/font/pdfs/5176.CFF.pdf">Subroutines</a>.
+
+Note: CFF glyph CharStrings can be flattened by inlining subroutines to become additional CharString bytes.
+
```
vlevantovsky, line 157: What if ... instead of including a list of things that _could be done_ to make a perfectly valid good (as in "efficiently designed") font file less efficient but usable with a range-request method, we could consider an alternative approach where we simply list range-request limitations and suggest that for TTF fonts that employ composite glyphs or for CFF fonts using subroutines other methods should be used - either a patch-subset or a full font download (that in many cases may be the most efficient option).

RangeRequest.bs
svgeesus: Interesting. Should probably be broken out into a separate issue

RangeRequest.bs
svgeesus: It isn't clear how range request can be applied to WOFF 1 either; it seems to involve compression and decompression to calculate the ranges.

RangeRequest.bs
svgeesus: At minimum this should be tested out.  Did we study the expansion from flattening and other pre-processing steps?

RangeRequest.bs
vlevantovsky: IIRC, the WOFF1 file was designed with the range request in mind. It starts with the Header and Table directory where each table is compressed separately, and the table directory provides an offset to the beginning of that table and the compressed length. So, while one cannot ask for a specific glyph range, downloading a set of e.g. metric/layout tables can be separated from the download of the glyph table data. And yes, implementing support for glyph ranges would involve compression and decompression of 'glyf' table.

BTW, on a separate issue, it seems that after the candidate corrections were published for WOFF2 - both [https://www.w3.org/TR/WOFF/](https://www.w3.org/TR/WOFF/) and [https://www.w3.org/TR/WOFF2/](https://www.w3.org/TR/WOFF2/) now point to the same updated WOFF2 Rec.

RangeRequest.bs
vlevantovsky: A while back, Monotype has made an effort to create a sizeable set of CJK fonts that would be suitable for OEM use (targeting the smallest storage size possible). They redesigned a [number of popular font families](https://catalog.monotype.com/font/oem-monotype/csong-tc/light-catt?keyword=CATT%20fonts) to employ as much of nested composite glyphs as possible, and released them under the CATT (Compact Asian TrueType) product name. IIRC, the size savings we've seen were quite good, e.g. an original flat design that needed ~9MB of storage would be reduced to ~4MB by utilizing composite glyph designs. I guess we can ask someone at Monotype to compare sizes of CATT fonts and their regular counterparts.

RangeRequest.bs
svgeesus: The  https://www.w3.org/TR/WOFF/  to  https://www.w3.org/TR/WOFF2/ redirect was reported and is now fixed

vlevantovsky: The [https://www.w3.org/TR/WOFF/](https://www.w3.org/TR/WOFF/) link still brings me to WOFF2 spec.

RangeRequest.bs
vlevantovsky: See issue #56 for continued discussion.

RangeRequest.bs
vlevantovsky: > As for data, the PFE analysis used flattened fonts for simulating the range request method so for the fonts tested by the analysis we saw significant byte reductions for CJK despite the flattening operation.

I do remember WG discussions related specifically to CFF (comparing the sizes of subroutinized vs. de-subroutinized charstrings). The size variations we saw were about 10-15%,  and de-subroutinization with subsequent Brotli compression produced better results in terms of bytes transferred overall. We did use flattened TTF glyph outlines for evaluation but I don't remember if we actually compared flattened fonts to their composite counterparts.

Considering the fact that simple and composite TTF glyphs are easily distinguishable, font optimization process can rearrange their sequences by re-grouping simple glyphs and composite glyphs together, to make them continuous data blocks for range-request mechanism. I honestly don't see an immediate extra benefit of flattening composite CJK fonts because preprocessing can easily analyze composite glyphs, group all simple radicals into a continuous data block, and transfer them in a single range request. Composite glyphs would then refer to these simple glyphs, and considering that heavily composited CJK fonts reuse many of these simple radicals many times over - I don't think that flattening TTF outlines is truly necessary for range-request transfer.

vlevantovsky: For continued discussion, see issue #58

RangeRequest.bs
litherum: Can we do this as a follow-up? If I move it now before the PR is landed, all the conversations on the PR would be marked as obsolete.

RangeRequest.bs
litherum: Because we have a dedicated issue for this, I'll resolve this for now, and add an issue in the PR with a link to the issue.

RangeRequest.bs
litherum: This appears to be what https://github.com/w3c/IFT/issues/59 is about, so I'll resolve this conversation and add an issue in the PR with a link to that issue.

RangeRequest.bs
```diff
@@ -10,26 +10,218 @@ Editor: Chris Lilley, W3C, https://svgees.us/, w3cid 1438
...
+<div class=example>
+The result of optimizing an OpenType font for the range-request IFT method is still a valid OpenType font.
+</div>
+
+Note: There are no <code>MUST</code>-level requirements on the organization of a [=range-request optimized font=]. Any arbitrary font file may be considered to be a [=range-request optimized font=]. However, additional optimizations should increase the performance of loading in a browser via the range-request IFT method. Font creators are encouraged to enact as many of the optimizations listed in this section as are reasonable for the fonts they create.
+
+<h3 id="font-organization-compression">
+Compression</h2>
+
+A [=range-request optimized font=] should not use any kind of compression other than [[!RFC7932]] (commonly referred to as "Brotli") compression.
```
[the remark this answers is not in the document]
litherum: I'd rather not mention any specific content-encoding algorithm; this IFT method is completely agnostic to it.

RangeRequest.bs
litherum: Now that I think about it, all the global functions should be in either `fpgm` or `prep`, which should come before `glyf`, so I suppose this constraint can just be deleted wholesale. 🆒

RangeRequest.bs
litherum: Because we have a dedicated issue for this, I'll resolve this for now, and add an issue in the PR with a link to the issue.

RangeRequest.bs
litherum: That choice is up to the user / font service / website author. This is the specification for the range request method; if someone reads this and concludes that they don't want to use the range request method, that's their choice.

RangeRequest.bs
```diff
@@ -10,26 +10,218 @@ Editor: Chris Lilley, W3C, https://svgees.us/, w3cid 1438
...
+
+The range-request method was designed to be as easy as possible to deploy, and doesn't require any server-side software. Deploying this method involves two steps:
+
+1. Modifying the CSS of the webpage to opt-in, by adding a particular keyword to particular places in the CSS file, and
+2. Optionally, rearranging the data inside the font file to optimize loading performance. This rearrangement process is a process by which a valid font file is input to the process and a valid font file is output from the process. When being served, the optimized font simply replaces the unoptimized font on the server. Every step in the optimization process is optional, but each should improve performance if performed. The optimization process is lossless, and no contours are affected.
+
+A rearranged font file is still a valid font file following existing formats; this specification does *not* describe a new font file format. The guidelines simply list strategies for organizing the data within the file to achieve the best performance with the range request method.
+
+<h2 id="opt-in">
```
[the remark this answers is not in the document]
litherum: Right. I believe this and the next section are both general to both methods. I've added some temporary text at the top of these sections for now, and when we merge the documents into one we can better organize the one shared document.

RangeRequest.bs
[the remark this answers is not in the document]
litherum: Okay, I've added a sentence here about these headers in this section, and I've also added a new top-level section called "Server Behaviors" which lists this information again, and also says that supporting range requests is required to support the range request IFT method.

RangeRequest.bs
vlevantovsky: Agree. However, making the spec more informative to aid to the understanding of all nuances that might affect this decision is a benefit, IMO. Since we already have a new issue #58 opened to address this topic, we should probably continue this discussion there.

svgeesus (approved):

PR #52, opened 2021-11-25
Make the link too fast hash a reference.

### LINKED ISSUE #44: Should fast hash be a normative reference?
Should [fast hash](https://github.com/ztanml/fast-hash) be a normative reference?

Or should it be an informative reference, with the pseudo-code in the spec the normative part?

(Either way it should be in the references)

garretrieger:
Fixes #44


svgeesus (approved):
LGTM

PR #54, opened 2021-11-26
Add preview and diff to spec pull requests

xfq:
Thank you for your work on this specification! I'm tracking the progress of this spec and hope it will improve page load time and make WebFonts more usable.

This PR adds [PR Preview](https://github.com/tobie/pr-preview) for the main spec. Although [PR Preview does not yet support multiple documents](https://github.com/tobie/pr-preview/issues/18), I think it is more convenient to have the preview and diff of the main document than no preview at all.

### DIFF
--- a/.pr-preview.json
+++ b/.pr-preview.json
@@ -1,0 +1,7 @@
+{
+    "src_file": "Overview.bs",
+    "type": "bikeshed",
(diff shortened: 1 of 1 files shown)


svgeesus (approved):
Thanks!

PR #76, opened 2022-02-06
Merging Patch-Subset and Range-Request specs

litherum:
This is just about the smallest kind of merge I could have performed. It places the shared sections at the front, and the Range Request sections at the end (in the existing section).

I didn't know how far to take @svgeesus's suggestion that we should organize the document from top-down rather than bottom-up, and I didn't want to make significant changes to the patch-subset part of the document. So that part is essentially unchanged.

I also deleted the existing sections about method selection, because I think the ones in the range request section are more comprehensively described. If you think this was a mistake, I'm happy to change this.

The other thing we might want to consider is beefing up the patch-subset section in the introduction, because it only has a single sentence about it.

In the future, we also might want to add a "so I'm a web author, which method should I be using?" section (or maybe even flowchart?).

***
<a href="https://pr-preview.s3.amazonaws.com/litherum/IFT/pull/76.html" title="Last updated on Feb 6, 2022, 5:11 AM UTC (c8a93d4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/76/a271e52...litherum:c8a93d4.html" title="Last updated on Feb 6, 2022, 5:11 AM UTC (c8a93d4)">Diff</a>


Overview.bs
```diff
@@ -91,22 +91,99 @@ Introduction {#intro}
 
 <em>This section is not normative.</em>
 
+Incremental Font Transfer (IFT) is a collection of technologies to improve the latency of remote fonts (or "web fonts") on the web. Without this technology, a browser needs to download every last byte of a font before it can render any characters using that font. IFT allows the browser to download only some of the bytes in the file, thereby decreasing the perceived latency between the time when a browser realizes it needs a font and when the necessary text can be rendered with that font.
+
 The success of WebFonts is unevenly distributed. This specification allows WebFonts to be used where
```
vlevantovsky, line 96: > The success of WebFonts is unevenly distributed.

Should we elaborate on this a bit further? For someone who haven't read PFE Evaluation Report, this statement may not seem obvious or self-explanatory.

Overview.bs
```diff
@@ -91,22 +91,99 @@ Introduction {#intro}
...
 slow networks, very large fonts, or complex subsetting requirements currently preclude their use. For
 example, even using WOFF 2 [[WOFF2]], fonts for CJK languages are too large to be practical.
 
+There are two different methods which can be used to incrementally transfer fonts.
+
+Patch Subset {#patch-subset}
+----------------------------
+
+The first method,
+<a href="patch-incxfer">Patch Subset</a>, uses a server which can generate binary updates to an existing font file.
```
vlevantovsky, line 106: > The first method, Patch Subset, uses a server which can generate binary updates to an existing font file.

Is this it, a complete description? I am sure there are many more details that we can and should provide about both methods, to help the first-time reader of the spec understand the IFT design and principles. As it's currently presented, all we are saying is that the first method is using a server, and the second method is very simple - hardly an accurate description.

Overview.bs
```diff
@@ -91,22 +91,99 @@ Introduction {#intro}
...
-so it is included in this specification as an alternative method.
+for many uses cases while still providing material improvments to loading performance for large fonts.
+
+Opt-In Mechanism {#opt-in}
+==========================
+
+<em>This section is general to both IFT methods.</em>
+
+The collection of IFT technologies utilizes a single shared opt-in mechanism. Each method does not use its own opt-in mechanism; instead, the webpage opts into the IFT technologies as a whole, and the browser and server negotiate to decide which specific method will be employed.
+
```
vlevantovsky, line 134: This section warrants additional discussions and may need to be revised. The opt-in mechanism is controlled by authors, including their preference of IFT method.

Overview.bs
garretrieger: as noted in the meeting today I can make a pass at fleshing this out after this PR gets merged.

garretrieger (approved):

vlevantovsky (approved):
Thank you, this is a good starting point for the development of the final recommendation-track document.

PR #77, opened 2022-02-10
Expand patch subset overview section. 

garretrieger:
For the most part moves text fom the later intro section up to the patch subset overview section.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/77.html" title="Last updated on Mar 8, 2022, 9:16 PM UTC (e729721)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/77/21043b6...e729721.html" title="Last updated on Mar 8, 2022, 9:16 PM UTC (e729721)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -117,7 +117,7 @@ Range Request {#range-request}
 ------------------------------
 
 The second
-method, <a href="#range-request-incxfer">Range Request</a>, has no server-side requirements other than the server should be able to respond to byte-based range requests. The browser makes range requests to the server for the specific bytes in the font file that it needs. In order to know which bytes are necessary, the browser makes one initial special request for the beginning of the file.
+method, <a href="#range-request-incxfer">Range Request</a>, has no server-side requirements other than the server should be able to respond to byte-based range requests. The browser makes range requests to the server for the specific bytes in the font file that it needs. In order to know which bytes are necessary, the browser makes one initial special request for the beginning of the file to obtain all required font tables, and then calculates glyph coverage and required byte ranges using font character-to-glyph mapping and glyph substitution / layout tables.
 
 In order for the range request method to be as effective as possible, the font file itself should be internally arranged in a particular way, in order to decrease the number of requests the browser needs to make. Therefore, it is expected that web developers wishing to use the range request method will use font files that have had their contents already arranged optimally.
 
--- a/Overview.bs
+++ b/Overview.bs
@@ -102,8 +102,17 @@ There are two different methods which can be used to incrementally transfer font
 Patch Subset {#patch-subset}
 ----------------------------
 
-The first method,
-<a href="patch-incxfer">Patch Subset</a>, uses a server which can generate binary updates to an existing font file.
+In the the first method, <a href="patch-incxfer">Patch Subset</a> a server generates binary patches which a
+client applies to a subset of the font in order to extend the coverage of that subset. The server
+is stateless, it does not maintain any session data for clients between requests. Thus when a client
+requests the generation of a patch from the server it has to fully describe the current subset of the
+font that it has in a way which allows the server to recreate it.
+
+Generic binary patch algorithms are used which do not need to be aware of the specifics of the font
+format. Typically a server will produce a patch by generating two font subsets: one which matches what
+the client currently has and one which matches the extended subset the client desires. A binary patch
+is then produced between the two subsets.
+
 
 Range Request {#range-request}
 ------------------------------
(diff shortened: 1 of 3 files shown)


Overview.bs
```diff
@@ -102,8 +102,17 @@ There are two different methods which can be used to incrementally transfer font
...
+client applies to a subset of the font in order to extend the coverage of that subset. The server
+is stateless, it does not maintain any session data for clients between requests. Thus when a client
+requests the generation of a patch from the server it has to fully describe the current subset of the
+font that it has in a way which allows the server to recreate it.
+
+Generic binary patch algorithms are used which do not need to be aware of the specifics of the font
+format. Typically a server will produce a patch by generating two font subsets: one which matches what
+the client currently has and one which matches the extended subset the client desires. A binary patch
+is then produced between the two subsets.
+
```
vlevantovsky, line 116: I'd like to avoid making statements in the spec that might constitute a judgement call. 
For example, stating that "Range Request is a very simple method of incremental transfer ..." is both misleading and a judgement call at the same time. I'd argue that while "Patch Subset" requires a more complex server-side support, it is simpler to implement on the UA side that already support WOFF2, Also, with Patch Subset a UA doesn't need to parse font data to figure out how to request and retrieve the next incremental update - one just asks for additional characters to be included. 

The Range Request method doesn't have server-side requirements and can be deployed on most of the servers without much effort, which is an advantage for some use cases, but implementations are definitely not going to be "very simple" - parsing partially downloaded font data to figure out glyph coverage and decide what byte range(s) needs to be requested (especially if complex scripts / layout tables are concerned) isn't something I'd call simple.

I'd suggest to make the overview a bit more descriptive to discuss why two different methods are offered and what's involved in the process of choosing one over another. In the end, it will be authors' decision which one to use, the UAs will need to be able to support both.

vlevantovsky (changes requested):

Overview.bs
svgeesus: I agree that making one part simpler has the trade off that other parts become more complex.

svgeesus (changes requested):
In general looks good. I agree with @vlevantovsky that the complexity trade-offs should be better described, rather than describing one method as "simple".  Range Request is simple server-side but has additional complexity client-side to compute the requested ranges.

garretrieger:
> In general looks good. I agree with @vlevantovsky that the complexity trade-offs should be better described, rather than describing one method as "simple". Range Request is simple server-side but has additional complexity client-side to compute the requested ranges.

I've removed the subjective statement about simplicity from the range request section. I wonder if we should add a third section after the patch subset and range request intro's that discusses the tradeoffs between the two?

vlevantovsky:
I like the new language in general, but I think we need to provide a bit more clarification on the mechanics of the range request method. In the very end of that paragraph, where the spec says:

> In order to know which bytes are necessary, the browser makes one initial special request for the beginning of the file.

I would suggest to extend that sentence to say:

"In order to know which bytes are necessary, the browser makes one initial special request for the beginning of the file to obtain all required font tables, and then calculates glyph coverage and required byte ranges using font character-to-glyph mapping and glyph substitution / layout tables."

One additional aspect of it (that probably deserves a separate issue and/or PR to be opened) is how much of this coverage calculation machinery has to be specified in details. It is easier for patch subset to just say that a server is responsible for calculating glyph coverage based on the requested character subsets, but I think we'd need to provide more details on what range request client implementation is expected to do w.r.t. glyph coverage calculations, taking into account glyph substitutions, stylistic sets, alternate forms, etc. Leaving it open-ended for implementers to guess isn't really an option, IMO.

garretrieger:
> In order to know which bytes are necessary, the browser makes one initial special request for the beginning of the file to obtain all required font tables, and then calculates glyph coverage and required byte ranges using font character-to-glyph mapping and glyph substitution / layout tables.

I added the suggested sentence, and agree that the second part of your comment probably warrants discussion in a separate issue.

garretrieger:
Adding Myles as a reviewer since some changes have been made to the range request overview section.

vlevantovsky:
> I added the suggested sentence, and agree that the second part of your comment probably warrants discussion in a separate issue.

Looks good, I will let @litherum review and comment. For additional discussion, see #79

Overview.bs
```diff
@@ -102,14 +102,22 @@ There are two different methods which can be used to incrementally transfer font
 Patch Subset {#patch-subset}
 ----------------------------
 
-The first method,
-<a href="patch-incxfer">Patch Subset</a>, uses a server which can generate binary updates to an existing font file.
+In the the first method, <a href="patch-incxfer">Patch Subset</a> a server generates binary patches which a
+client applies to a subset of the font in order to extend the coverage of that subset. The server
+is stateless, it does not maintain any session data for clients between requests. Thus when a client
```
svgeesus, line 107: Good, and important to note for the privacy folks who are concerned about state.

svgeesus (approved):

vlevantovsky (approved):
LGTM

svgeesus (approved):

PR #85, opened 2022-03-16
Update sparse bit set details.

garretrieger:
A couple of changes:
- Remove branch factor 16, and replace with branch factor 2. Our more recent simulations show branch factor 2 is useful to have.
- Change depth to be 5 bits which is sufficient to encode all 32 bit integers in all available branch factors. This leaves 1 reserved bit we can have for future use.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/85.html" title="Last updated on Mar 16, 2022, 12:37 AM UTC (8bab773)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/85/2a8c0d6...8bab773.html" title="Last updated on Mar 16, 2022, 12:37 AM UTC (8bab773)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -311,13 +311,13 @@ First append 2 bits which encode the branching factor:
   <tr>
     <th>Bits&nbsp;</th><th>Branching Factor</th>
   </tr>
-  <tr><td>00</td><td>4</td></tr>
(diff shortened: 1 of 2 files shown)


vlevantovsky (approved):
LGTM

PR #91, opened 2022-05-17
Fix #83

### LINKED ISSUE #83: Server may reply (how else can it, while still doing patch-subset)?
> If the server receives the patch-subset query parameter and wishes to honor it, the server may reply with a valid [patch subset response](https://w3c.github.io/IFT/Overview.html#PatchResponse)

So, it can also honor it with some other reply? How?

I can't see why this isn't a MUST. Or (to take into account that it wants to honor it but can't recreate the client state, a SHOULD)

svgeesus:
may -> should, with clarification for the case where range request unsupported. Prose and table now match :)

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -163,7 +163,7 @@ IFT Method Selection {#method-selection}
 When a page has indicated that a particular font should utilize IFT technology, the browser must determine which method to use. Different browsers may support different IFT methods, and different servers may support different IFT methods, so a negotation occurs as such:
 
 1. The browser makes the first request to the server. If the client prefers the <a href="https://w3c.github.io/IFT/Overview.html#patch-incxfer">patch-subset method</a>, it sends the relevant <a href="https://w3c.github.io/IFT/Overview.html#negotiating-transfer-type">query parameter</a>. If the client prefers the range-request method, it does not send the query parameter.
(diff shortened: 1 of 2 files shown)


vlevantovsky (approved):
"should" is definitely better then "may", so the proposed changes look fine. 
We may want to revisit "should" vs. "must" discussion - is there a way for server to signal specific condition that prevented recreating the client state? It seems like an additional roundtrip in this case is inevitable - can we help a client to "fix" patch subset request?

garretrieger:
> "should" is definitely better then "may", so the proposed changes look fine. We may want to revisit "should" vs. "must" discussion - is there a way for server to signal specific condition that prevented recreating the client state? It seems like an additional roundtrip in this case is inevitable - can we help a client to "fix" patch subset request?

As currently written (in https://w3c.github.io/IFT/Overview.html#handling-patch-request, specifically the note near the end of the section) if the server is unable to replicate the clients state (ie. checksums do not match up) it can respond with a "replacement" response which will instruct the client to replace it's current state with what the server sends. This allows the situation to be resolved without requiring an additional round trip.

svgeesus:
Per 17 May 2022 telcon, upgrade both of these to MUST

PR #92, opened 2022-05-17
fix #83 for realz

### LINKED ISSUE #83: Server may reply (how else can it, while still doing patch-subset)?

svgeesus:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/92.html" title="Last updated on May 17, 2022, 5:45 PM UTC (e6c7f57)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/92/affd4d6...e6c7f57.html" title="Last updated on May 17, 2022, 5:45 PM UTC (e6c7f57)">Diff</a>


vlevantovsky (approved):
Looks good.

PR #97, opened 2022-06-03
Re-work when original_font_checksum, original_features, original_axis_space, and codepoint_ordering are sent.

garretrieger:
Change to have these font derived fields sent whenever the server sees a mismatch between the client reported original font checksum and the servers original font checksum. This allows the fields to be updated in the client's state during a font version upgrade. Such as when a server is patching a client to a new version of the original font.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/97.html" title="Last updated on Jun 13, 2022, 9:26 PM UTC (aa8dc8f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/97/9fc60c1...aa8dc8f.html" title="Last updated on Jun 13, 2022, 9:26 PM UTC (aa8dc8f)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -804,7 +804,7 @@ For a PatchResponse object to be well formed:
      Only one of <code>patch</code> or <code>replacement</code> must be set.</span>
 *  <span class="conform server" id="conform-response-font-checksums">
      If either <code>patch</code> or <code>replacement</code> is set then <code>patch_format</code>,
-     <code>patched_checksum</code>, and <code>original_font_checksum</code> must be set.</span>
+     and <code>patched_checksum</code> must be set.</span>
 *  <span class="conform server" id="conform-response-valid-format">
      If <code>patch_format</code> is set then it must be one of the values listed in
      [[#patch-formats]].</span>
@@ -843,7 +843,8 @@ Additionally, the client can optionally store:
 *  Original font feature list: a list of the
     <a href="https://docs.microsoft.com/en-us/typography/opentype/spec/featuretags">opentype layout feature tags</a>
     that the original font has data for. This information can be used by the client to avoid sending
-    unnecessary requests for features which the original font does not contain.
+    unnecessary requests for features which the original font does not contain. Supplied by
+    <a href="#PatchResponse"><code>PatchResponse.original_features</code></a>.
 
 ### Extending the Font Subset ### {#extend-subset}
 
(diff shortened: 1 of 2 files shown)


Overview.bs
```diff
@@ -1081,36 +1085,40 @@ in an extended font subset:
...
     <code>codepoint_ordering</code> and <code>ordering_checksum</code> fields following
     [[#codepoint-reordering]].</span>
 
 *  <span class="conform server" id="conform-response-original-features">
-    If the <code>patch</code> or <code>replacement</code> fields are set then the
-    <code>original_features</code> field should be set to the list of feature tags that the original
-    font has data for.</span>
+    The response must set the <code>original_features</code> field to the list of
+    <a href="https://docs.microsoft.com/en-us/typography/opentype/spec/featuretags">opentype layout feature tags</a>
+    that the original font has data for.</span>
```
svgeesus, line 1104: what if the original font and the new, replacement font have different feature sets?

garretrieger: That's the motivation for this change. So how this works after this change is:
1. Say client has state for v1 of the font including original_feature list (v1).
2. At some point the server updates to v2 of the font which has a different feature list.
3. Client sends a patch request, server sees a mismatch between original checksum from the client and v2 of the font.
4. Server responds with  a patch that upgrades the clients v1 subset to the v2 subset with all features the client has previously asked for. This response will also include the original_features (v2) list (since it's mandatory to send when the checksum changes).
5. When client sees original_feature (v2) list included in the response it will replace the one previously saved list from v1 in its state.
6. Lastly, the client could check if there's features that it wants which are newly available in the v2 list and could potentially send an additional request to add them.

Overview.bs
```diff
@@ -1081,36 +1085,40 @@ in an extended font subset:
...
+Additionally:
+
+*  <span class="conform server" id="conform-response-patch-format">The format of the patch in the
+    either the <code>patch</code> or <code>replace</code> fields must be one of those listed in
+    <code>accept_patch_format</code></span>.
+
+*  <span class="conform server" id="conform-response-patched-checksum">
+     If <code>patch</code> or <code>replacement</code> fields are set the value of
+     <code>patched_checksum</code> must be set to the checksum of the extended font subset.
+     The checksum value must computed by the procedure in [[#computing-checksums]].</span>
```
vlevantovsky, line 1120: A minor change: The checksum value must _be_ computed

Overview.bs
```diff
@@ -962,10 +963,13 @@ should interpret and process the fields of the object as follows:
     resend the request that triggered this response but use the new codepoint ordering provided in
     this response.
 
-5. If <code>original_axis_space</code> is set then update the saved original axis space with the value
+5. If <code>original_features</code> is set and the client has opted to save them then update the saved
```
vlevantovsky, line 966: This list outlines the fields a client should process. In the edited step 5 description: 

> If <code>original_features</code> is set and the client has opted to save them then update the saved original feature list with the value specified in this field.

the last part of the sentence "then update the saved original feature list" is a bit ambiguous and may need to be clarified.

The rest of the changes LGTM.

Issue #108, opened 2022-07-08
Compression performance tradeoffs


yoavweiss:
Current Woff2 fonts are compressed with brotli 11 (typically), where the patch subsets are likely to be compressed with lower levels of brotli, to ensure on-the-fly compression for fast delivery. Hence, if we're expecting most of the font to be eventually downloaded, there's some cost to take into account (on top of extra requests, etc).

aside: [This research](https://dev.to/riknelix/fast-and-efficient-recompression-using-previous-compression-artifacts-47g5) may be applicable on that front and may reduce that cost, if large parts of the fonts are being downloaded.

It'd be good to clearly outline those downsides, in order to guide both user agents and developers on when they should or shouldn't be using patched incremental transfer.

The same principle also applies to range requests, which have many more downsides..

garretrieger:
> Current Woff2 fonts are compressed with brotli 11 (typically), where the patch subsets are likely to be compressed with lower levels of brotli, to ensure on-the-fly compression for fast delivery.

That's right, for example I believe in the demo we use brotli level 9.

> Hence, if we're expecting most of the font to be eventually downloaded

I don't think this is typically going to be true. To me using IFT makes the most sense if you have a font were you're expecting to not need all of the font (for a particular user) so for example fonts that support multiple languages but individual users are expected to only some user specific subset of those, large CJK fonts that contain a core set of frequently used characters and a really long tail of very infrequently used ones, or emoji/icon fonts.

Which gets at precisely what you mentioned that we should provide guidance in the spec on when it's appropriate to use incremental transfer and when it's better to just use a single file and/or multiple files via unicode range. I'll try writing up a new section non-normative section that gives some high level guidance and explains the trade offs being made.

svgeesus:
We should also spit the advice into alphabetic, shaping and CJK groups, like we did in the evaluagion report, because that affects how useful static subsetting it.

yisibl:
@svgeesus Are there any plans for WOFF3? In the future, is there a better compression algorithm to solve this problem?

svgeesus:
> Are there any plans for WOFF3? 

There are no concrete plans for WOFF3.

Issue #104, opened 2022-07-04
Explain more clearly why we have two methods and what the trade-offs are.


svgeesus:
One reviewer (from the IETF HTTP WG) [found no reason](https://lists.w3.org/Archives/Public/ietf-http-wg/2022AprJun/0279.html) for us to have two methods, and stated that it was impossible to choose between them.

> The most obvious question is about why you would provide *two* 
mechanisms. Nothing explores why you might prefer one over the other, at 
server or client. It is simply taken for granted there there are two. I 
don't know how a client might be expected to use this specification in a 
sensible fashion based on the information presented. Nor how a server 
might choose to deploy one or the other.

This needs to be made much clearer in the specification.

svgeesus:
TBH I thought this was [fully explained in the introduction](https://w3c.github.io/IFT/Overview.html#intro) but then, maybe this is the curse of knowledge if someone who is technically skilled but unfamiliar with the font work finds it poorly explained, we should fix it.

tiroj:
Section 1.3 of the introduction explains the basic difference of the two methods in terms of performance vs ease of deployment. That seemed adequate to me, but then I don’t actually manage font servers. Could it be fleshed out with some examples of when someone might choose one method over the other?

I also suggest editing the last paragraph of section 1 to read something like:

_There are two different methods which can be used to incrementally transfer fonts. The first method provides better performance while the second is considered easier to deploy._

Do we want to suggest that people should choose the Patch Subset method if they are able?

_____

Also note typo in the second paragraph of section 1.3:

_Range Request is simpler to deploy for many *uses* cases while still providing material improvments_

should be 

_use_

svgeesus:
@martinthomson do these recent changes address your concern?

martinthomson:
WFM

PR #112, opened 2022-07-19
Add "Performance Considerations" section.

garretrieger:
For #108 and #104.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/112.html" title="Last updated on Jul 19, 2022, 12:00 AM UTC (66cb400)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/112/aa09c4d...66cb400.html" title="Last updated on Jul 19, 2022, 12:00 AM UTC (66cb400)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -133,6 +133,80 @@ The evaluation report found that patch subset was generally more efficient in te
 performance and transferred bytes than range request. However, Range Request is simpler to deploy
 for many uses cases while still providing material improvments to loading performance for large fonts.
 
+Performance Considerations and the use of Incremental Font Transfer {#performance-considerations}
+--------------------------------------------------------
+
+Using incremental transfer may not always be beneficial, depending on the characteristics of the font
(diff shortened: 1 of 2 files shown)


svgeesus (approved):
Very nice summary, approve

skef:
I don't object to any of the existing content added in the current version of this PR, but it does seem to omit two topics: cache-ability (the likelihood that a given request can be offloaded to a caching service) and computational cost (which might be considered to be included under "latency" but the two can often be traded off). Any server host with significant load and less than unlimited resources may care about these issues.

garretrieger:
> I don't object to any of the existing content added in the current version of this PR, but it does seem to omit two topics: cache-ability (the likelihood that a given request can be offloaded to a caching service) and computational cost (which might be considered to be included under "latency" but the two can often be traded off). Any server host with significant load and less than unlimited resources may care about these issues.

Sounds good, I'll make a follow on set of changes to talk about these topics.

Issue #107, opened 2022-07-08
Negotiation algorithm is not ideal


yoavweiss:
Patch-subset-first requests have some costs, in the sense that they are optimistically uploading the patch.

It'd be better to have an [ALPS](https://www.ietf.org/archive/id/draft-vvv-httpbis-alps-01.html) frame that would opt in to patch-subset as part of the connection establishment to let the browser know of patching support ahead of request-sending time.

Squinting at this (a bit hard, admittedly), one can imagine patch-subsetting as some sort of a [Client Hint](https://github.com/WICG/client-hints-infrastructure/blob/main/README.md). It doesn't fully fit, as the server we want to opt-in to client hints here is not the top-level origin, but one of its font servers, and ALPS opt-in only works at the top-level origin.
But there are certainly parallels, and I can imagine y'all building something that is client-hint-like, with the right characteristics to solve this particular content negotiation problem.

I'm happy to advise on that front, if that's helpful.

garretrieger:
Thanks, I'm not too familiar with ALPS, so I'll review that and see if it looks like a better fit for doing the negotiation. It would be great if we could avoid sending the uneeded PatchRequest to a range request only server (so long as we're not adding any extra round trips to do so).

garretrieger:
I've reviewed ALPS and I have a couple concerns:
1. Since ALPS is part of TLS it works at the connection level. Connections are made to a specific domain and under that domain there may more than one actual servers processing requests depending on the specific patch requested. Having a frame in ALPS that opts in to patch subset would opt int all paths under that domain, which in many cases may not be ideal (eg. a CDN which serves assets for many different users from the same domain).
2. Using ALPS could complicate implementations where SSL termination is not happening at the same place where the HTTP request is being processed. Again thinking of the example where SSL termination is happening at a single place for a domain but specific paths under that domain are served by completely different servers.

I think then if we want to go down this path, we should make the use of a SETTING via ALPS optional. That is, define a server sent SETTING which instructs the client to only use range request. If the client does not receive that setting it will use the currently specified behaviour of including the PatchRequest message in the first request + fallback to range request depending on the server's response. If it does receive the setting then the client would omit the PatchRequest message from the first request. This then gives a mechanism to servers that don't support patch subset to have clients avoid sending them an unnecessary PatchRequest. 

The good news is it looks like we are free define custom setting identifier under HTTP3 (https://www.rfc-editor.org/rfc/rfc9114.html#name-defined-settings-parameters) as an extension.

[...]

Issue #75, opened 2022-02-06
Query parameters


mnot:
The IFT spec [describes](https://w3c.github.io/IFT/Overview.html#extend-subset) a format for use in query parameters of font URLs.

That's problematic, because it impinges on a [server's authority over its own URLs](https://www.rfc-editor.org/rfc/rfc8820.html).

The proper way to do this would be with request headers, probably [Client Hints](https://www.rfc-editor.org/rfc/rfc8942.html). If you don't want to do that, the next best thing would be to allow the site to dictate how to create a URL that contains the query information (e.g., using [URI Templates](https://www.rfc-editor.org/rfc/rfc6570.html)).

garretrieger:
To explain why we chose the query parameter:

- The most important requirement is that the initial request for a font should receive back the incremental data in only one round trip.
- In the initial request the client doesn't yet know if which variant of incremental transfer the server supports. Because we want the server to respond to the initial request with incremental data to satisfy the one round trip requirement we need to send enough information for either of the two server types to respond.
- Thus we need to ship some binary data (the PatchRequest message) with the first request.
- I considered a few alternative options to send that binary data:
  - Send the data in a header. Unfortunately this isn't an option because if we use any headers other than those that are cors safelisted there's the possibility of incurring a cors preflight request. Violating the one round trip requirement.
  - Send the first request as a POST request. This can be done without a CORS preflight, but isn't an option since a range request server won't necessarily be able to handle POST requests to static asset urls.
  - HTTP Methods other than GET/POST will incur cors preflight requests.

So that only leaves a query parameter on a GET request as a mechanism for passing the request data on the initial request.

Re: client hints, I don't think that helps us as client hints is about discovering client capabilities, where this is about discovering server capabilities and shipping down request data specific to incxfer.

To allow the content author to specify a URL format that would have to be part of the source url in the @font-face declaration. I'll have to look into that a bit more to see if that makes sense to do.

svgeesus:
On a related note, in the course of link fixing for the upcoming publication of the fully-merged IFT spec, we how hav e [defined term `query parameter`](https://w3c.github.io/IFT/Overview.html#query-parameter) in the [list of defined terms](https://w3c.github.io/IFT/Overview.html#index-defined-here).

@mnot the plan is to get network and perf review once that spec is published.

Oh and I also updated the HTTP [references](https://w3c.github.io/IFT/Overview.html#references) so we now cite RFC9110, RFC9111 and RFC9112

[...]

Issue #63, opened 2021-12-13
IFT Method Fallback table should have an "ambivalent client" column


litherum:
Right now there is a column in the [PR](https://github.com/w3c/IFT/pull/47) for a client which prefers range-request and a client which prefers patch-subset. There should also probably be a column for a client which wants to let the server decide for them.

garretrieger:
Fixed by #113

Issue #57, opened 2021-12-06
Supporting different IFT methods on client side


vlevantovsky:
In the [IFT spec](https://w3c.github.io/IFT/Overview.html#range-request-support) we mandate that a server supporting patch-request method must also support range-request method (which is a right thig to do, IMO). I couldn't find any language in the spec that would offer similar guidance for client implementations - I believe that in order to truly make it a viable option for authors the clients MUST support both IFT methods.

vlevantovsky:
To follow up on this based on the [WebFonts WG discussion](https://www.w3.org/2021/12/07-webfonts-minutes.html) - I agree that User Agents should always act on behalf of the user and enable the best user experience possible. If the user clicks on a link that brings him to a webpage that uses patch-subset method to optimize font load, or if a user has expressed the preference to use range-request whenever possible - the user agent shall honor that preference. In other words, allowing users and authors to have options to choose which IFT methods they want to use requires that User Agents must support both options.

litherum:
> we mandate that a server supporting patch-request method must also support range-request method

I think this is a different situation than your proposal here, because this existing mandate is necessary for clients to have the ability to choose which method they prefer. Specifically, this mandate removes a row from the table in the "IFT Method Fallback" section of the [PR](https://github.com/w3c/IFT/pull/47). If this row were not removed, method selection logic would have to handle this situation:

1. The client prefers range-request, but the server only supports patch-subset
2. The client, preferring range-request, sends a range-request to the server
3. The server, not supporting range-request, would reply with the whole font
4. The client at this point doesn't know whether the server is replying with the whole font because the server doesn't understand any IFT at all, or if the server is replying with the whole font because the server supports patch-subset but not range-request.
5. If the client wanted IFT (preferring range-request, but willing to fall-back to patch-subset), the client then would have to send a second request with the patch-subset information, completely discarding the information from the existing request/response
6. If the server doesn't understand any IFT, the server replies with the whole font _again_.
7. Now the client is downloading the whole font _twice_. This would be catastrophic.

Thus, the mandate on servers to support both methods is necessary for the method selection to work sensibly. Such a mandate on _clients_ is _not_ necessary.

It would be totally reasonable for a client implementation to want to implement (and ship) one of the methods before implementing the other, since neither is required for the other to work (from the client's perspective).

> the clients MUST support both IFT methods

Are you suggesting any kind of enforcement of this? Are you proposing that IFT servers would reject range-requests to font files?

[...]

PR #113, opened 2022-08-23
Revise method selection and negotiation.

garretrieger:
This introduces three css font tech keywords to aid method selection:
- incremental-patch: tells client to use patch subset
- incremental-range: tells client to use range request
- incremental-auto: tells client to use the existing method negotiation mechanism.

Additionally the negotiation mechanism has been revised to use a GET + Header in the initial request instead of GET + a URL query parameter. This addresses feedback about the use of a URL query parameter.

This is based on the findings of [Revisting Method Negotiation in Incremental Transfer](https://docs.google.com/document/d/1IcPg4o5k4OOmfLCfa1z2wQl6MD-kPBrahtcb4r-97ZU/edit?usp=sharing&resourcekey=0-9WtXKj_y4uihiB7K740xMQ)

Marking as draft, pending discussions with CSS wg about changing the incremental font tech keyword.

This will resolve the following issues:
#107 
#75 
#63 
#30 

and is relevant to some of the discussion in #57

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/113.html" title="Last updated on Aug 25, 2022, 9:23 PM UTC (b93ac7b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/113/7f379af...b93ac7b.html" title="Last updated on Aug 25, 2022, 9:23 PM UTC (b93ac7b)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -212,16 +212,24 @@ Opt-In Mechanism {#opt-in}
 
 <em>This section is general to both IFT methods.</em>
 
-The collection of IFT technologies utilizes a single shared opt-in mechanism. Each method does not use its own opt-in mechanism; instead, the webpage opts into the IFT technologies as a whole, and the browser and server negotiate to decide which specific method will be employed.
+Webpage's can choose to opt-in to either patch subset or range request incremental transfer for a font
+via the use of a CSS font tech keyword ([[css-fonts-4#font-tech-definitions]]) inside the
+''@font-face'' block.
 
-The opt-in mechanism is the <code>incremental</code> keyword inside the ''@font-face'' block. Websites specify this keyword inside their <code>@font-face</code> block, and the browser is then responsible for using IFT technologies, and for negotiating with the server to determine which specific IFT method to use.
+There are three tech keywords available:
+
+*  <code>incremental-patch</code>: Load the font incrementally using the patch subset method.
+*  <code>incremental-range</code>: Load the font incrementally using the range request method.
+*  <code>incremental-auto</code>:  Load the font incrementally. The particular method to use is
+    determined by <a href="#method-negotiation">Method Negotiation</a>.
 
 <div class=example>
-The use of the <code>incremental</code> keyword in this CSS rule indicates to the browser they should use IFT.
+The use of the <code>incremental-patch</code> keyword in this CSS rule indicates to the browser they
+should use the patch subset IFT method to load the font.
 <pre>
 @font-face {
     font-family: "MyCoolWebFont";
-    src: url("MyCoolWebFont.otf") tech(incremental);
+    src: url("MyCoolWebFont.otf") tech(incremental-patch);
 }
 </pre>
 </div>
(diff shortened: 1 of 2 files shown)


Overview.bs
```diff
@@ -212,39 +212,131 @@ Opt-In Mechanism {#opt-in}
...
 <em>This section is general to both IFT methods.</em>
 
-When a page has indicated that a particular font should utilize IFT technology, the browser must determine which method to use. Different browsers may support different IFT methods, and different servers may support different IFT methods, so a negotation occurs as such:
+The client selects the IFT method to utilize for a font load using the following procedure:
+
+1. If the content has specified which method to use, such as via the
+    <a href="#opt-in">opt-in mechanism</a>, then that method should be used if the client supports
+    it.
+
+2. If the client does not support the specified method then it should fallback to to loading
```
svgeesus, line 253: So (just to be clear) if the CSS specified `incremental-patch` and the client does not support it (but does support `incremental-range`); and the client has not contacted the server yet; then it must assume the server does not support `incremental-range` and just request the whole font? (And the same if the CSS says `incremental-range` and the client does not support it).

It seems odd to have this as step 2, and then step 3 is "if the client does not know the server capabilities" which implies at step 2 it *does* know.

What am I not getting, here?

garretrieger: I'll reword this. There are basically two cases:

content specifies incremental-patch or incremental-range which signals what the server supports (or in the non browser based case the client through some means knows which method the server supports). In that case proceed with the indicated method if supported.

Or content specifies incremental-auto (or this is a non browser based case where the server support is unknown in advance). Then the client should use negotiation to sort it out.

svgeesus: Got it.

svgeesus (approved):
This now looks good to me.

Suggest merging, so that the spec can be pointed to during CSSWG discussions

Overview.bs
```diff
@@ -212,39 +212,129 @@ Opt-In Mechanism {#opt-in}
...
-3. If the client receives the patch-subset magic number, it commences using the patch-subset method. Otherwise, if the client receives the <code>Accept-Ranges: bytes</code> header, it commences using the range-request method. Otherwise, the whole font file is downloaded, and the current non-incremental loading behavior is used.
+Note: the most recently specified method by the content takes precedence. For example if on a site the
+first page specifies to use the patch subset method for a font URL, but the second page specifies
+range request for the same font URL then the second page should load the font using range request.
+Such cases will require the client to reset any previously stored state for that URL and start fresh
```
vlevantovsky, line 261: Not sure if this Note is added to cover the hypothetical case of using two different methods on the same site and for the same font, but for sake of completeness, I guess we should also explain what should happen if the second page specifies <incremental-auto>. Possible scenarios would be to either renegotiate or to continue using previously specified <patch subset> method (which seems to be both logical and efficient from performance POV).

Overview.bs
```diff
@@ -212,39 +212,129 @@ Opt-In Mechanism {#opt-in}
...
+     using the range-request method. Otherwise, the whole font file is downloaded, and the current
+     non-incremental loading behavior is used.
 
 IFT Method Fallback {#fallback}
 -------------------------------
 
 <em>This section is not normative.</em>
 
-This summarizes behaviors that result from the above method selection.
+This summarizes behaviors that result from the above method selection and negotiation sections.
```
vlevantovsky, line 294: In the past, when we talked about whether we should mandate both methods, one of the arguments was that while it would be trivial for a patch-subset server to also support range-request, it may not be the case for a server that only supports range-request. The proposed method selection implies that either the author/content method selection is fulfilled, or the non-incremental approach is used, regardless of any other considerations. Should we revisit the earlier discussions to determine the possible outcomes?

Overview.bs
garretrieger: Agreed we should be able to just continue with the previous method in that case. I'll add clarification to this note.

Overview.bs
garretrieger: Patch subset servers supporting range request is still relevant to the incremental-auto case. For the method specific keywords (incremental-range/patch) I considered allowing incremental-patch to fallback to range if the client only supports range, but my thought is that the content has very explicitly asked for a specific method so we should only use the method that's been requested.

If the content is OK with either method it can declare a font face with two urls one tagged with incremental-patch and the other with incremental-range (or incremental-auto). Then it would be up to the client to pick which one to use.

garretrieger:
> This now looks good to me.
> 
> Suggest merging, so that the spec can be pointed to during CSSWG discussions

Agreed we could merge to make discussion with CSSWG easier.

Overview.bs
vlevantovsky: Makes sense. thank you for clarification.

vlevantovsky (approved):
Looks good, I agree we should merge to allow public discussions to happen during TPAC.

garretrieger:
FYI, opened an issue with the CSS WG: https://github.com/w3c/csswg-drafts/issues/7665

PR #128, opened 2022-11-09
Add a note about how feature selection interacts with IFT method selection.

### LINKED ISSUE #94: Redundant feature reduction
We might think about allowing and encouraging the client to specify what "level" of a feature it supports in cases where there may be redundant information in the font. 

I'm thinking of COLR v1 vs v0 vs BW fallback as the central example. A font may have all of these and there probably won't be overlap between the BW fallbacks and the v0/v1 glyphs. There may or may not be overlap between the v0 and v1 glyphs. The server implementation could be intelligent with COLR subsetting if it knows what the client can and will use (in a way that a "conventional" subsetter can't be).

(When well used CFF subroutinization and glyf composites may make such optimizations not worth the effort, although you'd still get the benefit of reducing the COLR table size. If COLR fonts become widely used they're probably the second best case for augmentation after CJK, though, so I think it's worth doing a bit of research on this possibility.)

garretrieger:
Fixes #94.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/128.html" title="Last updated on Nov 9, 2022, 7:47 PM UTC (ad1339f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/128/bc5b5fd...ad1339f.html" title="Last updated on Nov 9, 2022, 7:47 PM UTC (ad1339f)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -234,10 +234,19 @@ should use the patch subset IFT method to load the font.
 </pre>
 </div>
 
-Note: Each individual <code>@font-face</code> block may or may not opt-in to IFT. This is due to the variety of ways fonts are used on web pages. Authors have control over which fonts they want to use this technology with, and which they do not.
+Note: Each individual <code>@font-face</code> block may or may not opt-in to IFT. This is due to the
+variety of ways fonts are used on web pages. Authors have control over which fonts they want to use
+this technology with, and which they do not.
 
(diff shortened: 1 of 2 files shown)


svgeesus (approved):
Looks great

PR #132, opened 2022-11-17
Remove ordering and patched checksum from response.

### LINKED ISSUE #121: Questions about section 4.4: Client
I've been reading this section of the spec in light of the TPAC discussion and I have a few questions relating to client state, as specified in 4.4.1, and also just the protocol. If you want me to break this out into separate issues let me know.

1. There is an `accept_patch_format` field in the request but the guidance just says "set to the list of Patch Formats that this client is capable of decoding". Seems like if we have this field it should also be used to ensure the server sends a (genuine) patch that matches the previous subset or patch. So maybe once one patch has been downloaded the content of the field should be reduced to the type sent by the server the first time.
2. On the other hand, that may be overly constraining, as I can imagine wanting the subset and patch formats to be different. (We have this idea that "a subset is a patch of an empty file" but I'm not sure why we need to *enforce* that.) It seems like the server can easily distinguish between a subset request and a patch request based on the parameters (at least), so maybe this field should mean "what the client can decode" on the initial request and then "what the server sent last" on subsequent patch requests. 
3. The "font subset" of section 4.4.1 seems to have two roles. First, it is the file that gets "patched", and thus is the file that gets cached. Second, (I think) it is the file the client uses as a font. This strikes me as restrictive and maybe not how things tend to work in existing browsers for other formats.<p>For example, when a browser downloads a WOFF2 does it typically cache that file or the file extracted from it? (I would guess the former but I may be wrong.)<P>We talked about how you can't really "patch" a WOFF2 but the problems with doing so seem to relate to our particular model of what it means to patch something. If you can deterministically unpack an SFNT from a WOFF2 then you can patch it. Or maybe WOFF2 is unworkable but some similar format wouldn't be. <P>Wouldn't it be more general if the "font subset" part of the client state were just one of the valid "Patch Formats", which can then be unpacked into an SFNT (or whatever other end-use format)? Or, less specifically, wouldn't it be more general if what is cached need not be the end-use format?
4. Alternatively, if the end-use format *will* always be sole output representation of the patching process, why are we *requiring* a further checksum when SFNT (which is the only relevant underlying filetype for the foreseeable future) is already thoroughly check-summed? Are we saying that clients are relieved of checking those sums? (Do the clients not do those checks now?) Maybe to the extent that maybe you don't have to patch up `head` to have the right `checkSumAdjustment`? Or that the patch checksum and the SFNT checksums play different roles somehow? (Seems like we're proving our work over and over compared with, say, downloading a javascript file. Isn't this all TCP?)<P>If the cached file can be something other than the end-use font file then an intermediate checksum makes more sense to me.

garretrieger:
Both of these checksums are not strictly required. The ordering checksum can be computed by the client and patched checksum duplicates the purpose of checksums inside the font.

Fixes #121

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/132.html" title="Last updated on Nov 17, 2022, 11:27 PM UTC (7cf4ac6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/132/5d25f86...7cf4ac6.html" title="Last updated on Nov 17, 2022, 11:27 PM UTC (7cf4ac6)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -1040,18 +1040,15 @@ For a [=PatchRequest=] object to be well formed:
     <td><dfn for="PatchResponse">original_font_checksum</dfn></td><td>[=Integer=]</td></tr>
   <tr>
     <td>5</td>
-    <td><dfn for="PatchResponse">patched_checksum</dfn></td><td>[=Integer=]</td></tr>
-  <tr>
-    <td>6</td>
     <td><dfn for="PatchResponse">codepoint_ordering</dfn></td><td>[=IntegerList=]</td></tr>
   <tr>
-    <td>7</td>
+    <td>6</td>
     <td><dfn for="PatchResponse">subset_axis_space</dfn></td><td>[=AxisSpace=]</td></tr>
   <tr>
-    <td>8</td>
+    <td>7</td>
     <td><dfn for="PatchResponse">original_axis_space</dfn></td><td>[=AxisSpace=]</td></tr>
   <tr>
-    <td>9</td>
+    <td>8</td>
     <td><dfn for="PatchResponse">original_features</dfn></td><td>[=FeatureTagSet=]</td></tr>
 </table>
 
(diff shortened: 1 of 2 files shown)


svgeesus (approved):
Seems reasonable, good catch!

PR #135, opened 2022-11-30
Add small section about handling saving a page for offline usage.

### LINKED ISSUE #129: IFT and offline reading
Originally raised during charter review]():

@r12a said:

>  I wondered whether it's in scope to look at how the font transfer technology affects or is handled by offline reading of web pages, including saving pages from the browser. 

Does a local save count as a new origin or is it still attached to the origin from which it is downloaded? That affects what happens if the user traverses a link to another page on the original website while reading the saved version.

garretrieger:
Fixes #129


svgeesus (approved):
That seems to cover it, slightly sucks that any change triggers a full font download but there we are. Probably the most privacy-preserving option for what is now a local resource.

PR #138, opened 2023-03-10
Split off the range request section back into a separate document.

garretrieger:
This will allow the patch subset specification to be advanced without range request being complete.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/138.html" title="Last updated on Mar 13, 2023, 9:49 PM UTC (62e7c51)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/138/b9697c9...62e7c51.html" title="Last updated on Mar 13, 2023, 9:49 PM UTC (62e7c51)">Diff</a>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,10 @@
-all: Overview.html
+all: Overview.html RangeRequest.html
 
 Overview.html: Overview.bs feature-registry.html
 	bikeshed spec Overview.bs
+
+RangeRequest.html: RangeRequest.bs
+	bikeshed spec RangeRequest.bs
 
 feature-registry.html: feature-registry.csv registry_to_html.py
 	python3 registry_to_html.py > feature-registry.html
--- a/Overview.bs
+++ b/Overview.bs
@@ -1633,155 +1633,10 @@ file as a patch against a source file:
 Range Request Incremental Transfer {#range-request-incxfer}
 ===========================================================
 
-Introduction to Range Request {#range-request-intro}
------------------------------
-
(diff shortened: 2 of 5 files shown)


svgeesus (changes requested):
One typo, otherwise looks good

vlevantovsky (commented):
LGTM

skef:
I suggest holding off on this for the time being.

skef:
More specifically: I'm currently thinking we would be better ending up with either a single document or three documents. There's enough material (background, definitions, comparing the options) common to both that two seems like the wrong number. 

Of course, this is partly based on the prospects for replacing the range-request content with other related content.

svgeesus:
> What's the typo?

Thought I had added it as a review comment. And now can't see it. Nevermind.

svgeesus (approved):

skef:
> I would much rather merge this split, republish the documents, and start TAG review.

Start TAG review of both documents, or just the patch-subset spec? Do you think the range-request spec is ready to be reviewed?

skef:
I guess I have no inherent objections as long as substantial changes to the range-request spec (including almost complete replacement) and any further potential rearrangement (including re-integration of the specs, if desired) are still possible after the merge.

skef:
> In this regard, the range-request changes and future re-integration of a separate document, would definitely be something that is both possible and desirable.

In that case I withdraw any objections to merging this.

Issue #50, opened 2021-11-20
Proposal would allow pages to learn how the user is interacting with the site


pes10k:
This issue is part of the PING privacy review w3cping/privacy-request#61

The server (including third-parties) could observe the patch requests to learn how the user is interacting with the page.  In certain cases (e.g., when JavaScript is blocked) this would provide a unique powerful capability to the server, since they otherwise wouldn't be able to learn what the user is reading on the page, or typing in a form, etc.

I appreciate that this is partially discussed in Privacy Considerations section, but i think the analysis there is incorrect, or at least incomplete.  Specifically the text currently says: 

> One mitigation, which was originally introduced for reasons of networking efficiency so is likely to be implemented in practice, is to request additional, un-needed characters to dilute the ability to infer what content the user is viewing. Requesting characters which are statistically likely to occur may avoid a subsequent request.

In practice though, this is unlilkely to be a useful mitigation, at least in the way its currently discussed.  By definition, the client is likely to have already requested characters that are "statistically likely to occur" by the time the client needs to fetch an uncommon glyph / patch, meaning that there will be few or no common characters to request to obscure the uncommon, highly identifying characters.

I don't know what the best approach is here, but a partial solution here would be to define (or increase) the minimum patch size, reducing the granularity of the signal here.

svgeesus:
> The server (including third-parties) could observe the patch requests

How would third parties observe this, over HTTPS?

pes10k:
In several ways.  First, the font thats participating in IFT could be a third-party to the page no?

More generally though, users might disable JS to prevent the kind of moment-to-moment interaction monitoring that JS enables.  ITF would re-enable servers to learn (some) of the moment-to-moment, or keypress-by-keypress, monitoring the user was hoping to avoid by disabling JS.

Similarly, users of filter lists or similar tools (NoScript, etc) might disable scripts but not fonts, in an effort to match user privacy goals.  If a server is able to use ITF to learn about how an individual is interacting with a page, that increases the range of resources such tools need to restrict, and pushes users into a worse privacy/compatability tradeoff.

I appreciate that some of this can be done by maliciously constructed CSS as well, by constructing selectors that perform different network requests when different page elements appear or when text fields contain certain characters.  But since the target fonts here are enormous, with 10s of thousands of glyphs (or maybe more?) I expect there are margins where these attacks would be infeasable in CSS (where you had to enumerate every character targeted) but feasisble with IFT (where only a single instruction would suffice)

[...]

Issue #42, opened 2021-11-15
Repeat fingerprinting analysis assuming a malicious font server


samuelweiler:
The privacy analysis done to date seems to focus on "what fingerprinting could one do based on analyzing normal use patterns", e.g.:

> However, for those languages, it is possible that individual requests might be analyzed by a rogue font server to obtain intelligence about the type of content which is being read. It is unclear how feasible this attack is, or the computational complexity required to exploit it, unless the characters being requested are very unusual.

I would like the WG to analyze how IFT could be used for fingerprinting by malicious site with a self-controlled or colluding font server.  For example, if installed font state - which includes a set of characters in each font which the client might or might not have - is exposed in the same way across sites, could a site create a tool for "reading" that state vector by 1) specifying a self-controlled or colluding font server and 2) seeing what characters are loaded by a full array read?  Noting that the latter might be done invisibly to the user, using font color that matched the background or was otherwise not visible. 

This is just one example, which makes me think that the current analysis is incomplete.

svgeesus:
>  For example, if installed font state - which includes a set of characters in each font which the client might or might not have - is exposed in the same way across sites

No, see our response to [Require that incrementally-loaded fonts not be preserved nor exposed to other origins](https://github.com/w3c/IFT/issues/43). Because WebFonts must not be shared across origins, if site A uses font X and has n characters from that font, if I then go to site B which also uses font X it starts again with zero characters.

svgeesus:
@samuelweiler we responded a couple of months ago that the theoretical basis for the attack you suggested rests on an invalid premise, and is explicitly forbidden by the specification.

A malicious site or sites, therefore, even if colluding with a malicious or several malicious font servers, gets the same state information ("we have glyphs for zero characters") as any other site.

OK to close, therefore?

svgeesus:
@samuelweiler Noting that https://github.com/w3cping/tracking-issues/issues/263 still is tagged 'needs resolution' despite our arguing that your premise is invalid and specifically disallowed by the spec: please either close your issue or provide new evidence to convince us your criticism is valid. 

Thanks! (appreciate this is just probably lag in closing your issue)

svgeesus:
@samuelweiler ?

svgeesus:
[...]

PR #148, opened 2023-06-16
Add codepoint obfuscation via grouping to the patch subset specification.

garretrieger:
This adds a grouping mechanism to the patch subset client specification. This requires that for certain blocks of codepoints those codepoints always be requested in groups with a minimum size.

The impact of grouping on privacy was simulated here: https://github.com/w3c/PFE-analysis/blob/main/results/noise_simulations.md

There's still a couple of open questions we probably want to discuss with the WG before merging this:
1. What is an appropriate minimum group size? for now I've chosen 7 based on the simulation results.
2. Which codepoints specifically should we add to the list of codepoints to be grouped? Currently I only include all of the CJK ideograph blocks, but not any of the other CJK blocks (eg. punctuation, compatibility points, and so on)

Addresses: #42 and #50

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/148.html" title="Last updated on Jun 19, 2023, 8:26 PM UTC (4e6d7ae)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/148/058e0b8...4e6d7ae.html" title="Last updated on Jun 19, 2023, 8:26 PM UTC (4e6d7ae)">Diff</a>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,5 +9,8 @@ RangeRequest.html: RangeRequest.bs
 feature-registry.html: feature-registry.csv registry_to_html.py
 	python3 registry_to_html.py > feature-registry.html
 
+obfuscation-blocks.html: obfuscation_blocks_to_html.py
+	python3 obfuscation_blocks_to_html.py > obfuscation-blocks.html
+
 clean:
 	rm Overview.html RangeRequest.html feature-registry.html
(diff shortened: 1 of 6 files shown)


vlevantovsky (approved):
Looks good, thanks!

svgeesus (approved):

PR #149, opened 2023-06-20
Revamp the FeatureTagSet encoding scheme.

garretrieger:
'default included' features can now be omitted from the set by including the ID associated with that feature. This allows the client to fully customize the feature set if the list of default included features does not match what the client's shaper is expected to require for it's current content. For example the default included 'vert' feature could be omitted if no vertical layout is being done.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/149.html" title="Last updated on Sep 7, 2023, 5:17 PM UTC (77059cd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/149/9568adc...77059cd.html" title="Last updated on Sep 7, 2023, 5:17 PM UTC (77059cd)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -229,7 +229,7 @@ Opt-In Mechanism {#opt-in}
 
 <em>This section is general to both IFT methods.</em>
 
-Web page's can choose to opt-in to either patch subset or range request incremental transfer for a font
+Web pages can choose to opt-in to either patch subset or range request incremental transfer for a font
 via the use of a CSS font tech keyword ([[css-fonts-4#font-tech-definitions]]) inside the
 ''@font-face'' block.
 
--- a/Overview.bs
+++ b/Overview.bs
@@ -1865,9 +1865,24 @@ was assembled from several sources:
     default set in the <a href="https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-subset-input.cc">harfbuzz
     subsetter</a>.
 
-The entries in this table are versioned (by the Encoding Version column) since the client and server need to agree on
-the encoding. Currently there is only one version, 1, which is used by default. Future specification updates may
-introduce additional versions and a mechanism for the server to inform the client up to which version it supports.
+The columns should be interpreted as follows:
+
+*  Tag: the 4 byte tag for this feature from [[open-type/featurelist]].
+
+*  Encoded As: the integer ID used to represent this feature in a [=FeatureTagSet=].
+
+*  Encode If: entries marked as "not needed" are assumed to be in a [=FeatureTagSet=] by default. Including their
+    encoding ID in a [=FeatureTagSet=] removes the feature from the set. Entries marked as "needed" are not included
+    by default.
+    
+*  Encoding Version: the entries in this table are versioned (by the Encoding Version column) since the client and
+    server need to agree on the encoding. Currently there is only one version, 1, which is used by default. Future
+    specification updates may introduce additional versions and a mechanism for the server to inform the client up to
+    which version it supports.
+
+Note: the set of features selected to be included by default is purely an optimization mechanism and should not be
+used by clients as a guide for which features they should be asking for in requests. To minimize the number of bytes
+loaded clients should only request features that are needed for the rendering of the content used by the font.
 
 <pre class=include>
 <!-- Edit feature-registry.csv to update this table. -->
--- a/Overview.bs
+++ b/Overview.bs
@@ -864,27 +864,31 @@ bytes = [03 07 03 81 7F]
 ### <dfn>FeatureTagSet</dfn> ### {#featuretagset-object}
 
 A FeatureTagSet encodes a set of zero or more
-<a href="https://docs.microsoft.com/en-us/typography/opentype/spec/featuretags">OpenType layout feature tags</a>.
-Each feature tag is mapped to an integer value and then the set of mapped integers are encoded in a
-[=SortedIntegerList=]. Feature tags are mapped to integers as follows:
+<a href="https://docs.microsoft.com/en-us/typography/opentype/spec/featuretags">OpenType layout feature tags</a>. To encode
+a set of feature tags:
 
-*  If the tag is found in [[#feature-tag-list]]:
+1. Union the set of feature tags to be encoded and the list of feature tags in [[#feature-tag-list]]. For each tag in the
+    union:
 
-    *  If the "Encoded As" column corresponding to the tag is "default" then the tag is skipped and
-        not encoded.
+    *  If the tag is present in the list of feature tags to be encoded, is present in [[#feature-tag-list]], and
+        has a value of "needed" in the "Encode If" column of [[#feature-tag-list]]; then add the value of "Encoded As"
+        to the encoded ID list.
 
-    *  Else, the tag is mapped to the integer value in the "Encoded As" column.
+    *  If the tag is not present in the list of feature tags to be encoded, is present in [[#feature-tag-list]], and
+        has a value of "not needed" in the "Encode If" column of [[#feature-tag-list]]; then add the value of "Encoded As"
+        to the encoded ID list.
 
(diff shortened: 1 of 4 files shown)


skef:
When I was first looking at this the "needed"/"not needed" terminology confused me because I took it to refer to the listing rather than the feature it represents. My mistake, but maybe it's worth being a bit wordier.  

When I was first thinking about our decisions today I thought I would wind up recommending moving more of the "not needed" category to "needed". I talked myself out of that and I think got a better understanding of what we've arrived at. Accordingly, I think we should communicate something like the following between the section on forming the list and the appendix:

1. As the overall goal of the system is to reduce the amount of bytes transferred by omitting parts of the font that are not needed, it is best to arrive at the most accurate list of needed features as is practical.
2. A font shaper will sometimes make use of features that are not explicitly enabled, either because they are considered enabled by default (e.g. liga) or because a shaper queries the feature in certain contexts (e.g. numr). Therefore the accuracy of the feature list depends in part on the implementation of the shaper itself. Further, for this sort of contextual use, whether a feature will be queried may depend on which codepoints are to be rendered.
3. The assignment of numbered features to the two categories "needed" and "not needed" is based on research current practice in existing shapers. All features noted as "enabled by default" in the registry are included in the "not needed" category, as are any features that are used contextually (under some circumstance) according to (paper X).
4. The assignment of a feature to the "not needed" should not weigh against excluding it from a subset or augmentation (by adding to the list) if it is clear the feature will not be queried or activated. The registered features are split between the two categories so that a request with no non-explicit features specified (possibly because little is known about the shaper) is more likely to produce the same results as using the whole font. This aspect of the categorization should be considered advisory.

garretrieger:
> I think it would be worthwhile to also clarify normative / informative nature of references. For as long as we include the normative reference to the official ISO OFF standard, we can also use informative reference to publicly available OpenType spec (with the clarifying note that the content of it is identical to OFF), and then keep using OT links for convenience.

Done, I've updated the initial "Font Subset" section to reference the OFF standard and include a note that the remainder of the document references open type for convenience.

garretrieger:
> When I was first looking at this the "needed"/"not needed" terminology confused me because I took it to refer to the listing rather than the feature it represents. My mistake, but maybe it's worth being a bit wordier.
> 
> When I was first thinking about our decisions today I thought I would wind up recommending moving more of the "not needed" category to "needed". I talked myself out of that and I think got a better understanding of what we've arrived at. Accordingly, I think we should communicate something like the following between the section on forming the list and the appendix:
> 
> 1. As the overall goal of the system is to reduce the amount of bytes transferred by omitting parts of the font that are not needed, it is best to arrive at the most accurate list of needed features as is practical.
> 2. A font shaper will sometimes make use of features that are not explicitly enabled, either because they are considered enabled by default (e.g. liga) or because a shaper queries the feature in certain contexts (e.g. numr). Therefore the accuracy of the feature list depends in part on the implementation of the shaper itself. Further, for this sort of contextual use, whether a feature will be queried may depend on which codepoints are to be rendered.
> 3. The assignment of numbered features to the two categories "needed" and "not needed" is based on research current practice in existing shapers. All features noted as "enabled by default" in the registry are included in the "not needed" category, as are any features that are used contextually (under some circumstance) according to (paper X).
> 4. The assignment of a feature to the "not needed" should not weigh against excluding it from a subset or augmentation (by adding to the list) if it is clear the feature will not be queried or activated. The registered features are split between the two categories so that a request with no non-explicit features specified (possibly because little is known about the shaper) is more likely to produce the same results as using the whole font. This aspect of the categorization should be considered advisory.

Alright, took a swing at updating the appendix to better explain how the table works plus added a note to not rely on the default inclusion decision for the purposes of selecting features.

vlevantovsky:
> Done, I've updated the initial "Font Subset" section to reference the OFF standard and include a note that the remainder of the document references open type

We also need to update normative / informative references sections to include both ISO OFF and OpenType respectively.

garretrieger:
> > Done, I've updated the initial "Font Subset" section to reference the OFF standard and include a note that the remainder of the document references open type
> 
> We also need to update normative / informative references sections to include both ISO OFF and OpenType respectively.

Yeah, not sure why open-type is ending up in the normative section. It's only supposed to go there if we link to it with [[! .. ]]. I'll try and figure out what's going on.

Overview.bs
```diff
@@ -53,14 +53,13 @@ spec:fetch; type:dfn; for:/; text:response
     "date": "27 Jul 2021"
   },
 
-
-  "OpenType-Variations": {
-    "href": "https://docs.microsoft.com/en-us/typography/opentype/spec/otvaroverview",
+  "open-type": {
```
svgeesus, line 56: Specref already has OpenType:

[[OPENTYPE](https://www.specref.org/?q=opentype#)]

svgeesus:
> Yeah, not sure why open-type is ending up in the normative section. It's only supposed to go there if we link to it with [[! .. ]]. I'll try and figure out what's going on.

No, `[[! ]]` is an override to force what would have been an informative reference to be normative. All references in normative sections are considered normative. A Note: should be an informative section, though.

Overview.bs
garretrieger: Specifying our own version of the reference allows me to link to the specific sub pages. The spec ref one links to to default.htm so I can't do relative paths eg. [[open-type/otff#table-directory]]

garretrieger:
> > Yeah, not sure why open-type is ending up in the normative section. It's only supposed to go there if we link to it with [[! .. ]]. I'll try and figure out what's going on.
> 
> No, `[[! ]]` is an override to force what would have been an informative reference to be normative. All references in normative sections are considered normative. A Note: should be an informative section, though.

Ah got it, that makes more sense. Unfortunately with the way we're including the open type references directly along side normative text throughout it would be pretty difficult to move them all out to Note's.

garretrieger:
> > > Yeah, not sure why open-type is ending up in the normative section. It's only supposed to go there if we link to it with [[! .. ]]. I'll try and figure out what's going on.
> > 
> > 
> > No, `[[! ]]` is an override to force what would have been an informative reference to be normative. All references in normative sections are considered normative. A Note: should be an informative section, though.
> 
> Ah got it, that makes more sense. Unfortunately with the way we're including the open type references directly along side normative text throughout it would be pretty difficult to move them all out to Note's.

So if we really want the opentype references to be non-normative we may just need to reference only the OFF spec at the cost of the convenience of being able to link directly to relevant sections.

svgeesus:
> the convenience of being able to link directly to relevant sections

I think that is a significant benefit and would not like to see it go.

garretrieger:
> > the convenience of being able to link directly to relevant sections
> 
> I think that is a significant benefit and would not like to see it go.

At the last call we decided to leave this as is for now (w/ the open type references).

garretrieger:
This is ready for another look now. @vlevantovsky and @svgeesus could you take another look and approve it it looks good?

vlevantovsky:
Aside from the actual references to the spec source, I suggest removing "OpenType" references from the spec text (including one occurrence of the "open type" text, which is incorrect use of name). OpenType is a registered trademarked name, and should have a proper attribution when used. Repetitive use of the name in the IFT specification text does not really aid in anything, and seem redundant especially when it is used as part of the link to the relevant spec subclause (e.g. linking to "layout feature tags" would be just as effective as a link to "OpenType layout feature tags").

garretrieger:
> Aside from the actual references to the spec source, I suggest removing "OpenType" references from the spec text (including one occurrence of the "open type" text, which is incorrect use of name). OpenType is a registered trademarked name, and should have a proper attribution when used. Repetitive use of the name in the IFT specification text does not really aid in anything, and seem redundant especially when it is used as part of the link to the relevant spec subclause (e.g. linking to "layout feature tags" would be just as effective as a link to "OpenType layout feature tags").

Sounds good, fixed.

vlevantovsky (approved):
Looks good, thanks!

PR #154, opened 2023-11-24
Update references to RFC9110, RFC9111 and RFC9112

tidoust:
We eventually switched to the `https://httpwg.org/specs/` versions for RFC9110, RFC9111 and RFC9112 in the cross-references database, because they are more readable.

Unfortunately, this means that references to sections in these specs need to be updated again, as was done in #144. This update uses the new (and in theory now stable) fragment identifiers.

***
<a href="https://pr-preview.s3.amazonaws.com/tidoust/IFT/pull/154.html" title="Last updated on Nov 24, 2023, 3:15 PM UTC (643d5ab)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/154/01037d2...tidoust:643d5ab.html" title="Last updated on Nov 24, 2023, 3:15 PM UTC (643d5ab)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -311,14 +311,14 @@ When the particular IFT method(s) that are supported by a server are not known, 
 determine which method to use. Different clients may support different IFT methods, and different
 servers may support different IFT methods, so a negotiation occurs as such:
 
-1.  The browser makes the first request to the server using the GET HTTP method ([[rfc9110#section-9.3.1]]).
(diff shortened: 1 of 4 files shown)


tidoust:
Note that the underlying data has apparently not yet propagated to the [online Bikeshed service](https://api.csswg.org/bikeshed/), so PR Preview failed to generate the right links. This should happen soon, I just don't know when...

PR #155, opened 2024-02-08
Add a new static incremental font transfer spec to replace patch subset and range request.

garretrieger:
Replace patch subset spec with the the current static IFT draft. This draft is still in a very early stage, with missing sections and plenty of outstanding TODO's. However, I think it could be good to get it checked in to the main repository and iterate from there.

Static IFT is a new version of incremental transfer, that among other things allows for static hosting of patches. Like patch subset which came before it relies on patches to augment the font, but introduces a new mechanism for identifying and loading patches. Additionally it incorporates the glyph id based patching from the IFTB proposal (https://github.com/w3c/IFT/pull/151).

Couple of questions before merging:
1. Do the metadata fields at the top of Overview.bs need changing before merging (eg. the date, the WD status)?
2. For now I copied the previous patch subset spec text into PatchSubset.html, should we instead just remove it (and the RangeRequest one?)

***
<a href="https://pr-preview.s3.amazonaws.com/garretrieger/IFT/pull/155.html" title="Last updated on Feb 9, 2024, 11:03 PM UTC (26cbaaf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/155/e9c4e9c...garretrieger:26cbaaf.html" title="Last updated on Feb 9, 2024, 11:03 PM UTC (26cbaaf)">Diff</a>


svgeesus:
> 1. Do the metadata fields at the top of Overview.bs need changing before merging (eg. the date, the WD status)?

I would put

```html
Status: ED
Prepare for TR: no
```
and delete the Date: metadata, so it gets whatever the current date is.

svgeesus (approved):
I'm fine with merging this, makes it easier to work on the spec rather than on a PR.

PR #159, opened 2024-03-06
Add specification for the Format 2 Patch Map

garretrieger:
Adds spec text for the missing format 2 patch map section. Format 2 provides a general purpose mapping that supports design spaces and overlapping mapping entries.

Preview can be seen here: https://garretrieger.github.io/IFT/Overview.html#patch-map-format-2

***
<a href="https://pr-preview.s3.amazonaws.com/garretrieger/IFT/pull/159.html" title="Last updated on Mar 7, 2024, 10:52 PM UTC (9b8fd31)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/159/bec0db4...garretrieger:9b8fd31.html" title="Last updated on Mar 7, 2024, 10:52 PM UTC (9b8fd31)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -655,7 +655,8 @@ The algorithm:
     <td>uint8</td> <!-- TODO: uint32 to match format 1? -->
     <td><dfn for="Format 2 Patch Map">defaultPatchEncoding</dfn></td>
     <td>
-      Specifies the format of the patches linked to by uriTemplate. Using the ID number from the [[#font-patch-formats-summary]] table.
+      Specifies the format of the patches linked to by uriTemplate (unless overridden by an entry). Uses the ID numbers from the
+      [[#font-patch-formats-summary]] table.
     </td>
   </tr>
   <tr>
@@ -693,7 +694,7 @@ The algorithm:
   <tr>
     <td>[=Mapping Entry=]</td>
     <td><dfn for="Mapping Entries">entries</dfn>&lsqb;[=Format 2 Patch Map/entryCount=]&rsqb;</td>
-    <td>List of entries. Each entry has a variable length.</td>
+    <td>List of entries. Each entry has a variable length, which is determined following [$Interpret Format 2 Patch Map Entry$].</td>
   </tr>
 </table>
 
@@ -707,8 +708,8 @@ The algorithm:
     <td>uint8</td>
     <td><dfn for="Mapping Entry">format</dfn></td>
     <td>
-      A bit field. Bit 0 (least significant bit) through bit 5 indicate presence of optional fields. If bit 6 is set this entry is ignored.
-      Bit 7 is reserved for future use and set to 0.
+      A bit field. Bit 0 (least significant bit) through bit 5 indicate the presence of optional fields. If bit 6 is set this entry is
+      ignored. Bit 7 is reserved for future use and set to 0.
     </td>
   </tr>
 
@@ -723,7 +724,8 @@ The algorithm:
     <td>Tag</td>
     <td><dfn for="Mapping Entry">featureTags</dfn>[featureCount]</td>
     <td>
-      Array of [[open-type/featuretags|feature tags]]. Only present if [=Mapping Entry/format=] bit 0 is set.
+      List of [[open-type/featuretags|feature tags]] in the entry's [=font subset definition=]. Only present if
+      [=Mapping Entry/format=] bit 0 is set.
     </td>
   </tr>
 
@@ -738,7 +740,7 @@ The algorithm:
     <td>[=Design Space Segment=]</td>
     <td><dfn for="Mapping Entry">designSpaceSegments</dfn>[designSpaceCount]</td>
     <td>
-      Array of [=Design Space Segment=]'s. Only present if [=Mapping Entry/format=] bit 0 is set.
+      List of design space segments in the entry's [=font subset definition=]. Only present if [=Mapping Entry/format=] bit 0 is set.
     </td>
   </tr>
 
@@ -753,8 +755,9 @@ The algorithm:
     <td>uint16</td>
     <td><dfn for="Mapping Entry">copyIndices</dfn>[copyCount]</td>
     <td>
-      Array of indices from the [=Mapping Entries/entries=] array which should be copied into this mapping. All values must be less than
-      the index of this [=Mapping Entry=]. Only present if [=Mapping Entry/format=] bit 1 is set.
+      List of indices from the [=Mapping Entries/entries=] array whose [=font subset definition=] should be copied into this entry. All
+      values must be less than the index of this [=Mapping Entry=] in [=Mapping Entries/entries=]. Only present if
+      [=Mapping Entry/format=] bit 1 is set.
     </td>
   </tr>
 
@@ -762,7 +765,7 @@ The algorithm:
     <td>int16</td>
     <td><dfn for="Mapping Entry">entryIdDelta</dfn></td>
     <td>
-      Delta which is added to the entry id of the previous [=Mapping Entry=] to produce the entry id for this entry.
+      Signed delta which is added to the entry id of the previous [=Mapping Entry=] to produce the id for this entry.
       Only present if [=Mapping Entry/format=] bit 2 is set.
     </td>
   </tr>
@@ -771,7 +774,7 @@ The algorithm:
     <td>uint8</td>
     <td><dfn for="Mapping Entry">patchEncoding</dfn></td>
     <td>
-      Specifies the format of the patch linked to by this entry. Using the ID number from the [[#font-patch-formats-summary]] table.
+      Specifies the format of the patch linked to by this entry. Uses the ID numbers from the [[#font-patch-formats-summary]] table.
       Overrides [=Format 2 Patch Map/defaultPatchEncoding=].
       Only present if [=Mapping Entry/format=] bit 3 is set.
     </td>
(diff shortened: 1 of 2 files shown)


Overview.bs
```diff
@@ -623,13 +626,479 @@ The algorithm:
...
+  <tr>
+    <td>uint8</td>
+    <td><dfn for="Format 2 Patch Map">uriTemplate</dfn>[uriTemplateLength]</td>
+    <td>
+      A [[!UTF-8]] encoded string. Contains a [[#uri-templates]] which is used to produce URIs associated with each entry.
+    </td>
+  </tr>
+</table>
+
+<dfn>Mapping Entries</dfn> encoding:
```
skef, line 686: I'm not sure how I feel about these variable-length subtables being packed in this way, e.g. whether it might just make more sense to have offsets. But if we are going to do it this way, I think pretending they're records with this `entries[entryCount]` notation is misleading.  The OpenType spec would do something like "mappingEntries (variable)`.

One aspect of this is that this format seems like it could never be used directly -- you'll always have to unpack it into a separate representation in memory and then use that. Are we sure we're happy with that to save the extra bytes it would take to support direct use? Some of that will be regained by the compression. Is saving the cache space but paying for it in memory for the separate representation a false economy?

Overview.bs
```diff
@@ -623,13 +626,479 @@ The algorithm:
...
+
+<dfn>Mapping Entry</dfn> encoding:
+
+<table>
+  <tr>
+    <th>Type</th><th>Name</th><th>Description</th>
+  </tr>
+  <tr>
+    <td>uint8</td>
+    <td><dfn for="Mapping Entry">format</dfn></td>
```
skef, line 707: Why not "flags" instead of "format"?

Overview.bs
```diff
@@ -623,13 +626,479 @@ The algorithm:
...
+  <tr>
+    <td>Tag</td>
+    <td><dfn for="Design Space Segment">tag</dfn></td>
+    <td>
+      Axis tag value.
+    </td>
+  </tr>
+
+  <tr>
+    <td>Fixed</td>
```
skef, line 818: What's the point for Fixed for these? Why not F2DOT14? Is there some purpose for the increased granularity? Are you thinking these are in user units, so everything has to go through fvar and avar? That might be untenable when avar2 comes into the picture.

Overview.bs
```diff
@@ -623,13 +626,479 @@ The algorithm:
...
+    <td><dfn for="Mapping Entry">copyIndices</dfn>[copyCount]</td>
+    <td>
+      List of indices from the [=Mapping Entries/entries=] array whose [=font subset definition=] should be copied into this entry. All
+      values must be less than the index of this [=Mapping Entry=] in [=Mapping Entries/entries=]. Only present if
+      [=Mapping Entry/format=] bit 1 is set.
+    </td>
+  </tr>
+
+  <tr>
+    <td>int16</td>
```
skef, line 763: So is the idea here is that you can have an effective entry id greater than 2^16? Will that work with every needed combination? Do we need to push the compression that far? Or is it that you want sub-sequences of the records to stay consistent if there are earlier insertions? Given that the absolute entry IDs seem to have real meaning I'm not sure that would make sense.

Also, this description may not be entirely wrong given the vagueness of "produce" but it is misleading, given that the id will be [previous entry id] + [entryIdDelta] + 1. 

(Overall I'm asking why you can't just specify the entry ID with this field, making it be one plus the previous ID if the field is missing.)

skef (commented):
I took a quick look and I may have more thoughts later.

Overview.bs
garretrieger: Intention is to match the terminology used in GPOS ValueFormat which works similarly, but yeah this is probably a bit ambiguous with other uses of format numbers. How about formatFlags?

Overview.bs
garretrieger: Yes I've so far been treating this as being in the user units so I chose Fixed to match what fvar uses. The advantage to user units is the client/browser will be operating in user units and then won't need to know about or process internal font variation details to match mapping entries and apply patches. I'll see if I can add some text somewhere in the spec to make it explicit that throughout we're dealing in the external axis coordinates not the internal normalized ones.

I'm not super familiar with avar2, could you expand on what problems it might cause?

Overview.bs
garretrieger: Two reasons I see the delta being useful:
1. in the shared brotli case applying a patch will cause all of the remaining entries to get new id numbers further down in the space, so the delta can be used to jump the first entry id ahead and the rest will increment by one as normal.
2. We have the ability to have mapping entries that are copied into others but set the ignore bit so they themselves aren't used. Deltas could be used to keep the non-ignored entries having sequential ids. Though thinking about this, maybe I should specify that ignored entries don't have entry ids and thus don't cause the normal +1 increment.

I don't see a need to have >2^16 id's, but there's also no reason they wouldn't work as things currently are. Though we probably should set an upper bound somewhere (like how I disallow negative values).

garretrieger: I'll clarify the description to include the "+1". That's further clarified later on as well in the interpreting section, but might as well keep this consistent.

Overview.bs
skef: formatFlags would work.

Overview.bs
skef: Makes sense.

Overview.bs
skef: Hmm. I suppose one can always include a superset of the axis information if one needs it. 

I recommend chatting a bit with Behdad a bit about this aspect of the project. He'll be able to give opinions on whether avar2 could cause problems.

Overview.bs
skef: I think omitting the IDs for ignored entries makes sense.

> in the shared brotli case applying a patch will cause all of the remaining entries to get new id numbers further down in the space, so the delta can be used to jump the first entry id ahead and the rest will increment by one as normal.

So the idea is that given that the segmentation is fixed across the whole graph, portions of the map for a given part of the segmentation will tend to repeat at different nodes on the graph, and the delta helps with that? I can see how that will work in practice with the implicit + 1 when the field is missing, I'm more fuzzy on it working when you need to specify the field in the first place. I guess the idea is that what a patch does is remove an entry, and the delta at that point will need to change but the other deltas already established from earlier patches won't need to?

Overview.bs
skef: The short answer, from what I understand and remember, is that the mapping through avar 2 can be quite complex, leading to gaps. But presumably the subsetter would just *include* those gaps, so maybe it isn't a big deal.

Overview.bs
garretrieger: Changed to be uint8 entries[variable].

Overview.bs
garretrieger: Sounds good, I'll follow up with Behdad. My take is that this is will likely be a problem for subsetter implementations to figure out (supersetting to cover gaps seems reasonable to me) and if we stick to user axis space here, it hopefully won't be something we specifically need to address in this spec.

Overview.bs
garretrieger: On second thought we'll want to keep assigning IDs to ignored. When applying glyph keyed patches we want there to be a mechanism where the client can remove entries without modifying the length of the IFT/IFTX tables. The ignored bit allows us to do that in format 2, but only if we keep assigning ignored entries IDs (so ids for following entries aren't changed).

To give a concrete example if we start with a subset which has patches 1...n. After applying one of those patches all patches 1...n no longer can be applied since the base has changed, and thus mapping will be updated to instead link to a new set of patches with ids m+1 ... m + n - 1 (where m depends on the patch applied). The key side (subset def) of the mapping entries remains the same, but all of the ID's need to be shifted. The first entry get's a delta of +m, the applied entry is removed, and all remaining entries are left untouched and use the implicit +1.

Also depending on how an encoder is implemented it may choose for some reason not to use completely sequential id's. Having the deltas allows for arbitrary ID's sequences to be assigned to each entry so the encoder isn't restricted in choosing ids.

svgeesus (approved):
Looks good overall.

Fine with merging once @skef comments are addressed.

Overview.bs
garretrieger: > (Overall I'm asking why you can't just specify the entry ID with this field, making it be one plus the previous ID if the field is missing.)

On re-reading this I think I missed this part. Setting the id value directly should also work fine for the ways I typically see this being used. For most cases I believe the encoded size would be the same using deltas or absolute values.

I still have a slight preference for using deltas as it does work nicely in a case where you have a non-sequential list ID's that you want to shift uniformly with a patch application (the current encoder prototype doesn't make use of this, but maybe others might?).

I've also been thinking we probably want to increase the size of this field to at least 24 bits (whether it's a delta or absolute value) and the entry count to 32 bits to match format 1.

skef: I think that question from the first response is resolved for me. As I understand it the idea of using the delta instead of the value is that subsequent patches to earlier portions of the range will need to change the delta in fewer cases.

Overview.bs
garretrieger: To follow up on this Behdad has started to do some work on planning out how subsetting might work for avar2. Don't have a general solution yet, but sounds like we should be able to get to something workable for most cases. After some more though we'll definitely need to stick to user space units on the IFT side since if the font is initially static (with option to augment design space) then the client won't have any information on the mapping from user space to normalized space.

Overview.bs
garretrieger: Going to merge this as is for now to unblock some follow on changes, but I'll follow up with another PR to adjust the field sizes, and if needed we can continue the discussion on absolute vs delta ids there.

PR #161, opened 2024-04-05
Updates to the encoder section + TODO cleanup

garretrieger:
- Further clarifies some of the text in the encoder section.
- Adds more of the text for encoder guidance.
- Cleans up a bunch of TODO's throughout.
- Change from "shared brotli" to "brotli"

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/161.html" title="Last updated on Apr 12, 2024, 10:11 PM UTC (d90c786)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/161/545a4a1...d90c786.html" title="Last updated on Apr 12, 2024, 10:11 PM UTC (d90c786)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -1148,13 +1148,13 @@ The following patch formats are defined by this specification:
     <th>Description</th>
   </tr>
   <tr>
-    <td>[[#shared-brotli]]</td>
+    <td>[[#brotli]]</td>
     <td>[=dependent=]</td>
     <td>1</td>
     <td>A brotli encoded binary diff that uses the entire [=font subset=] as a base.</td>
   </tr>
   <tr>
-    <td>[[#per-table-shared-brotli]]</td>
+    <td>[[#per-table-brotli]]</td>
     <td>[=dependent=]</td>
     <td>2</td>
     <td>A collection of brotli encoded binary diffs that use tables from the [=font subset=] as bases.</td>
@@ -1169,7 +1169,7 @@ The following patch formats are defined by this specification:
 
 More detailed descriptions of each algorithm can be found in the following sections.
 
-Brotli Patch {#shared-brotli}
+Brotli Patch {#brotli}
 ------------------------------------
 
 In a brotli patch the target file is encoded with [[!RFC7932|brotli compression]] using the
@@ -1202,7 +1202,7 @@ of a short header followed by brotli encoded data.
   </tr>
 </table>
 
-<h4 algorithm id="apply-shared-brotli">Applying Brotli Patches</h4>
+<h4 algorithm id="apply-brotli">Applying Brotli Patches</h4>
 
 This algorithm is used to apply a brotli patch to extend a [=font subset=] to cover additional codepoints,
 features, and/or design-variation space.
@@ -1235,7 +1235,7 @@ The algorithm:
     using the <var>base font subset</var> as a [[Shared-Brotli#section-3.2|shared LZ77 dictionary]]. Return the decoded
     result as the <var>extended font subset</var>
 
-Per Table Brotli {#per-table-shared-brotli}
+Per Table Brotli {#per-table-brotli}
 --------------------------------------------------
 
 A per table brotli patch contains a collection of patches which are applied to the individual
@@ -1301,7 +1301,7 @@ of that [=TablePatch=].
 </table>
 
 
-<h4 algorithm id="apply-per-table-shared-brotli">Applying Per Table Brotli Patches</h4>
+<h4 algorithm id="apply-per-table-brotli">Applying Per Table Brotli Patches</h4>
 
 This algorithm is used to apply a per table brotli patch to extend a [=font subset=] to cover additional codepoints,
 features, and/or design-variation space.
@@ -1668,7 +1668,10 @@ The [=incremental font=] and associated patches produced by a compliant encoder:
      render identically to the fully expanded font for that content.
 
 4.  When an encoder is used to transform an existing font into an [=incremental font=] the associated
-     [$Fully Expand a Font Subset|fully expanded font$] should be equal to the existing font.
+     [$Fully Expand a Font Subset|fully expanded font$] should be equivalent to the existing font. An equivalent fully expanded font
+     should have all of the same [[open-type/otff#table-directory|tables]] as the existing font (excluding the incremental IFT/IFTX
+     tables) and each of those tables should be functionally equivalent to the corresponding table in the existing font. Note: the fully
+     expanded may not always be an exact binary match with the existing font.
 
 When an encoder is used to transform an existing font file into and [=incremental font=] and a client is implemented according to the
 other sections of this document, the intent of the IFT specification is that appearance and behavior of the font in the client will be the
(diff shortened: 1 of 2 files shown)


svgeesus (approved):
All looks good to me

Overview.bs
```diff
@@ -1688,11 +1688,55 @@ guidance that encoder implementations may want to consider.
...
+
+Example pseudo code for a very basic encoder implementation using a font subsetter:
+
+```
+# Encodes a font (full_font) into an incremental font that starts at base_subset_def
+# and can incrementally add any of subset_definitions. Returns the IFT encoded font
+# and set of associated patches.
+encode_as_ift(full_font, base_subset_def, subset_definitions):
+  base_font = subset(full_font, base_subset_def)
+  patches = encode_node(full_font, base_font, base_subset_def, subset_definitions)
```
skef, line 1709: You could make this

```
patches, base_font = encode_node(full_font, base_font, base_subset_def, subset_definitions)
```

to make it clearer that the base_font is modified by encode_node

Overview.bs
```diff
@@ -1688,11 +1688,55 @@ guidance that encoder implementations may want to consider.
...
+encode_node(full_font, base_font, cur_def, subset_definitions):
+  patches = []
+  next_fonts = []
+  
+  for each subset_def in subset_definitions not fully covered by cur_def:
+    next_def = subset_def union cur_def
+    next_font = subset(full_font, next_def)
+    let patch_url be a new unique url
+    
+    add patch map entry into base_font from (subset_def - cur_def) to patch_url
```
skef, line 1723: I first misread this thinking "to" was associated with "add", implying `patch_url` was being changed. Might something like `add a mapping from (subset_def - cur_def) to patch_url into the patch map` be clearer?

Overview.bs
```diff
@@ -1688,11 +1688,55 @@ guidance that encoder implementations may want to consider.
...
+    patches += encode(full_font, next_font, next_def, subset_definitions)
+
+    next_fonts += (next_font, next_def, patch_url)
+
+  for each (next_font, next_def, patch_url):
```
skef, line 1728: Let's make this

```
for each (next_font, next_def, patch_url) in next_fonts:
```

to be more explicit.

Overview.bs
```diff
@@ -1688,11 +1688,55 @@ guidance that encoder implementations may want to consider.
...
+  patches = []
+  next_fonts = []
+  
+  for each subset_def in subset_definitions not fully covered by cur_def:
+    next_def = subset_def union cur_def
+    next_font = subset(full_font, next_def)
+    let patch_url be a new unique url
+    
+    add patch map entry into base_font from (subset_def - cur_def) to patch_url
+    patches += encode(full_font, next_font, next_def, subset_definitions)
```
skef, line 1724: Is this line left over from an earlier edit? I don't understand how this patches line relates to the one at 1730.

Overview.bs
```diff
@@ -1688,11 +1688,55 @@ guidance that encoder implementations may want to consider.
...
+  for each (next_font, next_def, patch_url):
+    patch = dependent_patch_diff(base_font, next_font)
+    patches += (patch, patch_url)
+  
+  return patches
+```
 
-<!-- TODO: describe a simple encoder which utilizes only a subsetter -->
+In this example implementation if the union of the input base subset definition and the list of subset definitions fully covers the input
+full font, and the subsetter implementation used correctly retains all functionality then the above implementation should meet the
```
skef, line 1736: comma before "then"

Overview.bs
```diff
@@ -349,21 +340,23 @@ Some example inputs and the corresponding expansions:
...
 The 'IFTX' table is optional. When both tables are present, the mapping of the font as a whole is the union of the mappings of
-the two tables.
+the two tables. The two new tables are used only in this specification and are not being added to the [[open-type|Open-Type]]
+specification.
 
 Note: allowing the mapping to be split between two distinct tables allows an incremental font to more easily make use of multiple
 patch types. For example all patches of one type can be specified in the 'IFT ' table, and all patches of a second type in the
 'IFTX' table. Those patches can make updates only to one of the mapping tables and avoid making conflicting updates.
 
+Note: [=incremental font|incremental fonts=] can be converted to [[WOFF2]] and still be utilized as an incremental fonts. In this
```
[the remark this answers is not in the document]
garretrieger: It's my current understanding that the woff2 glyf transformation is predictable (while it will change the original glyf table, it will always produce the same result when decoded regardless of decoder implementation), so the encoder should be able to generate patches against the transformed and then decoded glyf table. Which is what my current prototype encoder does.

However, I'll need to spend some time reviewing the woff2 spec to make sure that is indeed a correct assumption. If I find that isn't true then I'll update the guidance here to recommend not using the glyf/hmtx transforms in woff2. Also I think I'll add a recommendation to only use per-table and/or glyph keyed patches  so that you don't have to rely on the woff2 decoders table directory and table ordering.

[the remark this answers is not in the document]
skef: If you verify that I think the section will need to be more specific, and say something about either WOFF2-encoding and then decoding the subset font prior to generating the patch or otherwise isolating how the format will change.

Also, just as we are with this spec we need to be careful about what the WOFF2 spec mandates versus suggests. Supposing the glyf transformation is predictable is that a reliable property of the spec or just of the reference implementation? If it's the latter we would need to say something about knowing how your specific encoder works (or maybe in that case it would just be better to punt).

Overview.bs
garretrieger: Rewrote and also added some commas to hopefully make it more clear.

Overview.bs
garretrieger: Done.

Overview.bs
garretrieger: This recursive call generates all of the patches for the lower levels of the graph while on line 1730 patches is further extended to add all of the patches associated with the current level. So encode(..., next_font,.. ) will generate any patches from next_font to anything else reachable from it, and line 1730 will generate any patches from base_font to any of the next_font's.

Overview.bs
garretrieger: Done.

Overview.bs
garretrieger: Done.

Overview.bs
```diff
@@ -1688,11 +1688,55 @@ guidance that encoder implementations may want to consider.
 
 <b>Utilize an existing font subsetter implementation</b>
 
-<!-- TODO: A high quality encoder should preserve functionality as discussed above. Existing font subsetters can be used to assist
-     constructing encodings that meet this requirement. They provide all of the logic needed to preserve functionality. High quality
-     open source subsetters exist -->
+As discussed in [[#encoder]] a high quality encoder should preserve the functionality of the original font. One way to
+achieve this is to leverage an existing font subsetter implementation to produce font subsets that retain the functionality
+of the original font. The IFT patches can then be derived from these subsets.
```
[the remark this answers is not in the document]
garretrieger: Done.

Overview.bs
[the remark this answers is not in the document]
garretrieger: Agreed, on both. I'll update here once I've had a chance to review the woff2 spec in more detail.

Overview.bs
skef: In that case what is the form or type of a "patch"? In (what was) 1730 a patch is a tuple that includes a `patch_url` but in this line neither `patch_url` nor `base_font` are passed to `encode`, so it doesn't seem likely it would be returning a tuple that includes that value.

garretrieger: Oh sorry for the confusion, I forgot to rename "encode(" to "encode_fonts(" (just pushed a fix). This is meant to be a recursive call so the return is list of tuples of the (binary patch, url). In the base case where encode_fonts(...) is generating the patches to the leaf nodes line 1730 will create the initial set of tuples, and then each recursive call above will add additional tuples for it's specific layer in the graph.

Overview.bs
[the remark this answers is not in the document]
garretrieger: After a review of the woff2 spec here's my conclusions:
1. The woff2 decoder does not gaurantee a specific table ordering in the decoded file.
2. For transformed glyf/loca the woff2 decoder does not gaurantee a predictable binary encoding of the reconstructed glyf table. The transformed glyf table decodes to a list of x/y points which the decoder is free to choose how to represent in the glyf format (eg. choosing encoded widths for the coordinate values).
3. Transformed hmtx tables should have predictable decoded results.

Based on these here's what I'm proposing as guidance we give, when using IFT + a WOFF2 encoded font:
1. The IFT encoding should not use "Brotli Patches". These cannot be safely applied to an unknown table ordering resulting from the woff2 decoding.
2. The IFT encoding should not use "Per Table Brotli Patches" that modify glyf/loca IF the WOFF2 encoded contains a transformed glyf/loca table. The decoded glyf/loca is not predictable and cannot be safefy patched via a brotli patch. Glyph-keyed patches may be safely combined with a transformed glyf/loca table.
3. The IFT encoding can use "Per Table Brotli Patches" that modify transformed hmtx tables, however, it's recommended they do not out of an abundance of caution.
4. All IFT decoding/processing must happen after woff2 decoding is applied.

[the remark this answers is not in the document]
garretrieger: Just added a new section which captures this guidance.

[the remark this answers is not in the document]
garretrieger: Going to mark as resolved, feel free to reopen if you have additional comments about the WOFF2 guidance text.

PR #162, opened 2024-04-09
Add text for the "Load a Patch File" algorithm.

garretrieger:
This defines patch URI's as URI references which may be relative. Relative references use the containing incremental font URI as the base URI.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/162.html" title="Last updated on Apr 9, 2024, 10:38 PM UTC (5ee2cea)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/162/25b4515...5ee2cea.html" title="Last updated on Apr 9, 2024, 10:38 PM UTC (5ee2cea)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -1592,21 +1592,27 @@ The algorithm:
 
 <dfn abstract-op>Load patch file</dfn>
 
-<!-- TODO: when discussing how to load patches via URI's we need to define what they are relative too -->
-<!-- TODO: ref fetch spec like we did in the patch subset spec. -->
 <!-- TODO: consider requiring HTTPS (or disallowing HTTP specifically if we want to allow file:// and other url types) -->
 
 The inputs to this algorithm are:
 
-*   <var>URI</var>: a URI identifying the patch file to load.
+*   <var>Patch URI</var>: A [[rfc3986#section-4.1|URI Reference]] identifying the patch file to load. As a URI reference this may be a
+     relative path.
+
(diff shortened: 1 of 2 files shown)


skef (approved):
This looks good to me.

I think at some past meeting there were questions about whether the patch URIs should be relative to something else, either optionally or always, but I'm not aware of a compelling reason to make them relative to anything but the initial download URL, so I continue to favor this approach.

On the protocol restrictions subject: Is there any reason to have restrictions on the patches that aren't on the initial font file? I can't think of any offhand. And since the URIs would be relative it seems like that reasoning would take care of the problem.

garretrieger:
> > On the protocol restrictions subject: Is there any reason to have restrictions on the patches that aren't on the initial font file? I can't think of any offhand. And since the URIs would be relative it seems like that reasoning would take care of the problem.
> 
> I'm currently specifying the use of [URI References](https://datatracker.ietf.org/doc/html/rfc3986#section-4.1) for the patches. These can be absolute or relative. So as things currently are if we do decide to introduce protocol restrictions we would need to do so for both the initial font file and the patches.
> 
> I can envision some cases where you might have a different protocol (scheme in the URI spec) on the patches than the init file. For example, you could theoretically embed the initial font file as a data URI in the CSS while having the patches loaded with absolute URI's to remove a round trip (assuming your init file is fairly small).

As for the restrictions, agreed that'd we'd likely want the same restrictions in place for both init font and patches, but just make sure to do it in a way that allows the two to have different protocols.

PR #163, opened 2024-04-10
Fill out the 'Handle errors' section.

garretrieger:
Allow extension to optionally continue in the case of a failure to load a URL, otherwise recommend extension stop.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/163.html" title="Last updated on May 23, 2024, 9:08 PM UTC (38d3371)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/163/062767e...38d3371.html" title="Last updated on May 23, 2024, 9:08 PM UTC (38d3371)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -474,12 +474,16 @@ The algorithm:
 
 <dfn abstract-op>Handle errors</dfn>
 
-Coming soon.
+If the extending the font subset process has failed with an error then, some of the data within the font may not be fully loaded and as
+a result rendering content which relies on the missing data may result in incorrect renderings. The client may choose to continue using
+the font, but should only use it for the rendering of codepoints, features, and design space that are fully loaded. An element of the
+font can be considered fully loaded if data for that element is present in the font (eg. for codepoints, listed in the cmap) and there are
+no mapping entries in the font subset which reference that element. Rendering of all other content should fallback to a different font
+following normal client fallback logic.
 
-<!-- TODO -->
-
-<!-- TODO: we should consider defining a valid encoded IFT font as one that produces a correct font subset for the corresponding
-           definition in the entry. This would likely be the only requirement made for encoders. -->
+If the error occurred during [$Load patch file$] then, the client may continue trying to extend the font subset
+if there are remaining patches available other than the one(s) that failed to load. In the case of all other errors the client
+must not attempt to further extend the font subset.
 
 <h3 algorithm id="fully-expanding-a-font">Fully Expanding a Font</h3>
 
--- a/Overview.html
+++ b/Overview.html
@@ -7,7 +7,7 @@
   <link href="https://www.w3.org/StyleSheets/TR/2021/W3C-ED" rel="stylesheet">
   <meta content="Bikeshed version d765c696b, updated Fri Mar 8 15:58:52 2024 -0800" name="generator">
   <link href="https://www.w3.org/TR/IFT/" rel="canonical">
-  <meta content="bcb93fd5c909171e15f96d6ee03abc89685c2b51" name="revision">
+  <meta content="062767e81b858e014d16c80241f8886229c5e759" name="revision">
   <meta content="dark light" name="color-scheme">
   <link href="https://www.w3.org/StyleSheets/TR/2021/dark.css" media="(prefers-color-scheme: dark)" rel="stylesheet" type="text/css">
 <style>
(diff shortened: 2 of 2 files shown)


Overview.bs
```diff
@@ -1610,12 +1610,11 @@ The algorithm:
...
 <dfn abstract-op>Handle errors</dfn>
 
-Coming soon.
-
-<!-- TODO -->
-
-<!-- TODO: we should consider defining a valid encoded IFT font as one that produces a correct font subset for the corresponding
-           definition in the entry. This would likely be the only requirement made for encoders. -->
+If the extending the font subset process has failed with an error, the client can continue to use the current extended font subset
+(from prior to the error occuring) to render the content, while falling back to the another font to render parts of the content not
```
skef, line 1614: I don't disagree with any of this but it seems tricky to get right in practice so I'm wondering if this is a bit understated. Are you thinking a client might make a copy with a modified cmap? Or maybe (in effect) put the right combination of unicode-range notations in place around it?

And is there a definitive means of determining what code points the initial file or a given patch level supports? (I suppose there probably is, I'm just worried about the "can-provide-more-than-what-was requested" cases.)

What if the failure had was trying to add an axis, or range of an axis? What if it was adding a feature?

garretrieger: Yeah you'e right, I wasn't considering the case where the cmap overdeclares coverage that is expected to be filled in later via patches. I can see two approaches that we could take for the error handling:

1. Recommend a font that has suffered a failure during extension not be used at all, this is the safest most conservative option.
2. Allow the font to be used after failure but only for codepoints that are listed in the cmap and are not listed in any of the patch entries that are not yet loaded (ie. so any codepoints that we are reasonably confident do not have any pending loads). There of course could still be missing functionality w/ features and or design space, we could add some additional text cautioning against using a partially loaded font in those cases.

What do you think?

skef: Either of these options seem OK to me. i suppose the decision should rest on whether the cost of communicating (and comprehending) the more liberal option is outweighed by the benefit. I doubt it would be in the short term because it would be annoying to implement correctly, but specifications aren't just for the short term.

garretrieger: I took a pass at rewriting this. I decided to present both options and leave it up to the client to decide which to use. I can see both being valid options for a client depending on the specifics of the client.

Overview.bs
```diff
@@ -1646,12 +1646,17 @@ The algorithm:
...
-<!-- TODO -->
-
-<!-- TODO: we should consider defining a valid encoded IFT font as one that produces a correct font subset for the corresponding
-           definition in the entry. This would likely be the only requirement made for encoders. -->
+If the extending the font subset process has failed with an error then, some of the data within the font may not be fully loaded and as
+a result rendering content which relies on the missing data may result in incorrect renderings. The client may choose to continue using
+the font, but should only use it for the rendering of codepoints, features, and design space that are fully loaded. An element of the
+font can be considered fully loaded if data for that element is present in the font (eg. for codepoints, listed in the cmap) and there are
+no mapping entries in the font subset which reference that element. Rendering of all other content should fallback to a different font
+following normal client fallback logic. Alternatively, after a failure the client may choose to treat the entire font as failed and not
```
svgeesus, line 1654: That gives more flexibility but reduced testability. What is the advantage of rejecting the entire font? What class of error is avoided by doing so?

Overview.bs
```diff
@@ -1646,12 +1646,17 @@ The algorithm:
...
+use it at all, falling back to the next font.
+
+If the error occured during [$Load patch file$] then, the client may continue trying to extend the font subset
+if there are remaining patches available other than the one(s) that failed to load. In the case of all other errors the client
+should not attempt to further extend the font subset.
```
svgeesus, line 1659: I think this case could be a MUST

garretrieger: That seems reasonable, I'll change it.

Overview.bs
garretrieger: What I intended here is to give a spectrum of options to a client implementation for how to handle failures and let them choose what makes sense given client specific details. However, I think given that the first option is specified as a "should" and makes sense for the vast majority of clients then, we can probably omit specifically calling out the option to reject the whole font. It's technically still allowed if there's a client implementation where it makes sense.

garretrieger: Alright, removed that sentence.

Overview.bs
garretrieger: Done.

svgeesus (approved):
Describing both options does seem clearer.

PR #166, opened 2024-04-13
Rework patch categorization

garretrieger:
For context see: https://lists.w3.org/Archives/Public/public-webfonts-wg/2024Apr/0000.html

This is the set of associated spec changes for the above proposal.

In summary:
- Replaces dependent/independent patch categorizations with Full/Partial/No Invalidation.
- Make it explicit that mapping table unique ID's are used to track patch invalidation.
- Update the extension algorithm to work with the new categorizations.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/166.html" title="Last updated on Apr 24, 2024, 11:37 PM UTC (efa0908)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/166/d225b17...efa0908.html" title="Last updated on Apr 24, 2024, 11:37 PM UTC (efa0908)">Diff</a>


svgeesus (approved):
This three-way categorization of ID-based invalidation seems a lot clearer to me than the older dependent/independent one.

PR #171, opened 2024-04-29
Major Reorganization of Spec Sections.

garretrieger:
This reorders several of the top level sections in order to make the specification easier to understand by an unfamiliar reader:
1. Move the 'Extending a Font Subset' section to before the patch map and patch formats section. This introduces the process which utilizes patch maps and patches before getting into the minute details of those.
2. In the common definitions section, add basic summaries of patch maps and patches. This is needed to introduce the basics on these concept before they are used in the 'Extending a Font Subset' algorithm.
3. Move the 'URI Templates' from the common definitions into the patch map table section, which is the only section that references it.
4. Move the 'Patch Invalidations' section from the patch map section and into the 'Extending a Font Subset Section' where it's needed to understand how the extension algorithm works.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/171.html" title="Last updated on Apr 30, 2024, 11:57 PM UTC (2cd6eb0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/171/7d7db59...2cd6eb0.html" title="Last updated on Apr 30, 2024, 11:57 PM UTC (2cd6eb0)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -331,7 +331,7 @@ codepoints, layout features and/or design space.
 Patch Invalidations {#font-patch-invalidations}
 -----------------------------------------------
 
-The application of a patch to an an incremental [=font subset=] may invalidate some or all of the other mapped patches, so that if
+The application of a patch to an incremental [=font subset=] may invalidate some or all of the other mapped patches, so that if
 one of those were to be applied the result would be a corrupt
 [=font subset|font=]. Patch validity is tracked by the compatibility ID from the [[#patch-map-table]]. Every patch has a
 compatibility ID encoded within it which needs to match the compatibility ID from the [[#patch-map-table]] which lists that patch.
--- a/Overview.bs
+++ b/Overview.bs
@@ -316,8 +316,8 @@ in [[#font-format-extensions]].
   </tr>
 </table>
 
-Data Types {#data-types}
-------------------------
+Explanation of Data Types {#data-types}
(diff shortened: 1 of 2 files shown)


Overview.bs
```diff
@@ -316,25 +316,26 @@ in [[#font-format-extensions]].
...
-This sections defines the algorithm that a client uses to extend an [=incremental font|incremental=] [=font subset=] to cover additional
+This section defines the algorithm that a client uses to extend an [=incremental font|incremental=] [=font subset=] to cover additional
 codepoints, layout features and/or design space.
 
 Patch Invalidations {#font-patch-invalidations}
 -----------------------------------------------
 
-Patches in an incremental [=font subset=] when applied may invalidate other available patches, making them invalid to be applied
-to the extended [=font subset=]. Patch validity is checked using the compatibility ID from the [[#patch-map-table]]. Every patch has a
+The application of a patch to an an incremental [=font subset=] may invalidate some or all of the other mapped patches, so that if
```
skef, line 334: I guess I left an extra "an" here.

skef (approved):
Other than my stray "an" this looks good to me, and is a definite improvement on the sorts of things I've been worried about.

Overview.bs
garretrieger: Fixed

PR #175, opened 2024-05-17
First pass at adding iteration limits to the extension algorithm.

garretrieger:
I've introduced two mechanisms to limit the number of iterations that can be made in the extension algorithm (to prevent malformed/bad fonts from causing an excessive or infinite number of iterations):

1. Added text to prevent re-loading the same patch url multiple times (this prevents a case where a patch that does nothing is applied over and over again).
2. Introduced iteration limits:

- Limited the number of partial/full invalidation patch loads to 100, this limit is lower because these patches have to be loaded and applied serially.
- Limited the total number of patch applications of any type to 2000. This limit is higher since No Invalidation patches will loaded and applied in parallel.

The specific numeric limits are a first pass and I'm open to discussion on what would be good values for these limits.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/175.html" title="Last updated on May 28, 2024, 5:52 PM UTC (5103182)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/175/7479aca...5103182.html" title="Last updated on May 28, 2024, 5:52 PM UTC (5103182)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -385,9 +385,12 @@ The algorithm:
     <var>target subset definition</var> as inputs, if it returns false remove <var>entry</var>
     from <var>entry list</var>.
 
-5.  If <var>entry list</var> is empty, then the extension operation is finished, return <var>extended font subset</var>.
+5.  Remove any entries in <var>entry list</var> which have a patch URI which was loaded and applied previously during the execution
+    of this algorithm.
 
-6.  Pick one <var>entry</var> from <var>entry list</var> with the following procedure:
+6.  If <var>entry list</var> is empty, then the extension operation is finished, return <var>extended font subset</var>.
+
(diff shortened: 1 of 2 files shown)


svgeesus (approved):
Useful to have these limits to prevent Denial of Service attacks from carefully-crafted fonts. So this should perhaps be called out in the Security section? But that can easily be a separate PR, no need to hold this one up.

garretrieger:
> Useful to have these limits to prevent Denial of Service attacks from carefully-crafted fonts. So this should perhaps be called out in the Security section? But that can easily be a separate PR, no need to hold this one up.

Good idea, that would definitely be good to discuss there.

skef (approved):
(Setting aside than the suspiciously base-10-looking limit numbers) this looks good to me

PR #186, opened 2024-06-18
Add more guidance about network round trips + patch types.

garretrieger:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/186.html" title="Last updated on Jun 19, 2024, 5:08 PM UTC (22f8626)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/186/97c7aee...22f8626.html" title="Last updated on Jun 19, 2024, 5:08 PM UTC (22f8626)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -2012,7 +2012,8 @@ are typically downloaded and applied in series, which has implications for the p
 [[#glyph-keyed]] patches are quite distinct from the other patch types. First, [[#glyph-keyed]] patches can only modify 
 tables containing glyph outline data, and therefore an [=incremental font=] that only uses [[#glyph-keyed]] must include all
 other font table data in the initial font file. Second, [[#glyph-keyed]] patches are [=No Invalidation|not Invalidating=],
-and can therefore be downloaded and applied independently.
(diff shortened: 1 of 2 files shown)


Overview.bs
```diff
@@ -2042,6 +2048,17 @@ way:
 4. Lastly, use glyph keyed patches to update the remaining tables, here much smaller fine-grained segments can be utilized without
     requiring too many patches.
 
+The mixed patch type encoding is expected to be the more commonly used approach since many fonts will fall somewhere in between the two
```
skef, line 2051: The more I think about the construction "the more commonly" the less feeling I have of what level of emphasis I'm supposed to understand from it. We could say " the most commonly" and that would be clear, or we could say "a more commonly" which is vague but self-consistently vague. 

The description "many fonts" seems like it captures the desired (?) vagueness on its own, however, so how about just "... is expected to be a commonly used approach ..."

skef (approved):
LGTM (other than my usage quibble)

Overview.bs
garretrieger: Updated to "is expected to be a commonly used approach"

PR #187, opened 2024-07-02
initial explainer rewrite

### LINKED ISSUE #160: Rewrite explainer
The [current explainer](https://github.com/w3c/IFT/blob/main/Patch-Subset-Explainer.md) will need to be totally re-written. There is no longer a need to explain why we have patch-subset and range-request; but there is now a need to explain why there are two patch formats, for example.

svgeesus:
Would close #160


IFT-Explainer.md
```diff
@@ -0,0 +1,161 @@
...
+However, instead of requiring custom patch generation for each user,
+the initial font has two new [_Patch Map_ tables](https://w3c.github.io/IFT/Overview.html#patch-map-dfn)
+which map each subset definition to **the link** of the relevant patch.
+Links are stored as RFC 6570 [URI Templates](https://www.rfc-editor.org/rfc/rfc6570).
+
+Thus, static hosting of files (base fonts, and patches) is easy,
+and cache performance is good
+because patches are shared between users.
+
+Both **independent** (commutative) and **dependent** (non-commutative)
```
skef, line 143: Even in a document intended to be this high-level I think we should clarify here that the spec supports "**independent** (commutative)" patches for glyph-only data and "**dependent** (non-commutative)" patches for other data. (Technically the latter can patch both but I think at this level that can be elided.) The current spec achieves a compromise between factors you've just discussed and the specific patch types are the essence of that compromise.

IFT-Explainer.md
```diff
@@ -0,0 +1,161 @@
...
+patches are [supported](https://w3c.github.io/IFT/Overview.html#font-format-extensions).
+
+In addition, patches can add or extend design axes
```
skef, line 146: This is true but because of the two patch types maybe a bit less true of this system than the last one (in that to **add** an axis you'll typically wind up downloading some duplicate glyph data). 

@garretrieger is this OK or should we soft-petal it more?

skef (commented):
As a top-level comment, I wonder if, structurally speaking, we've said enough about "**independent**" patches. In recent conversations (I wasn't entirely sure what each of us was thinking earlier) I think we've established that the concept-formerly-known-as-IFTB is what allows the new spec to address the general IFT problem. (That is, you may not always use independent/glyph-keyed patches, but without them the current spec wouldn't be performant in some common cases, including typical CJK cases.) And IFTB wasn't derived directly from Patch-Subset (or Range-Request). 

However, "**independent**" patches just sort of appear in the discussion without much said or even implied about their role in the overall solution. 

Of course, we don't need or want to say anything very substantial about IFTB in such a high-level document, at most a sentence or two. That or those would express:

1.  We incorporated a new idea to allow patches of glyph data to be independent,
2. This type of patch would have been too costly to compute dynamically but they work well with this new framework of pre-computed patches. 

That seems like enough to explain the change. @svgeesus think you can boil that down to match everything else?

IFT-Explainer.md
```diff
@@ -0,0 +1,161 @@
...
+by validating the request,
+computing a binary patch between the current, subsetted font on the client
+and the desired subset of the original font,
+and then sending the patch,
+which the client applied to produce a new, enlarged subset font.
+
+It therefore required new server capabilities,
+in addition to client changes.
+
+Work on Range Request was discontinued because, despite
```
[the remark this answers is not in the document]
svgeesus: Yes, oops!

IFT-Explainer.md
```diff
@@ -0,0 +1,161 @@
...
+[median byte reductions of 90% for large fonts](https://www.w3.org/TR/PFE-evaluation/#analysis-cjk),
+
+- The dynamic subsetting severely impacted CDN **cache performance**
+- There was no subsetting of _design axes_, an unacceptable result with **variable fonts** becoming ubiquitous
+- Subsetting fonts with **complex interactions between glyphs** was challenging
+- Non-outline data was not subsetted, and can be significant (e.g. Color fonts)
```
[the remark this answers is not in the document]
svgeesus: Right, I was going on this from the slideset:

> Limitations of Current IFTB Proposal
> ● Designed to only augment outline data. For many fonts this makes sense but
for some types of families non-outline data can be significant. (eg. complex
GSUB/GPOS, COLRv1)

I agree that deletion is the best way to resolve it

IFT-Explainer.md
```diff
@@ -0,0 +1,161 @@
...
+
+For fonts with _many glyphs_ (such as are typically used for Chinese and Japanese, for example),
+even with the compression provided by WOFF 1 or 2,
+download sizes are still **far too large**
+with a median WOFF2 size of [**1.8MB**](https://www.w3.org/TR/PFE-evaluation/#font-langtype).
+Thus, usage of Web Fonts in China and Japan is [**close to zero**](https://www.w3.org/TR/PFE-evaluation/#fail-large).
+
+For languages with a small set of glyphs, **static font subsetting** is widely deployed.
+
+However, for those languages with **complex shaping requirements**,
```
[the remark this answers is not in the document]
svgeesus: fixed

IFT-Explainer.md
```diff
@@ -0,0 +1,174 @@
...
+in addition to client changes.
+
+Work on Patch Subset was discontinued because, despite
+[median byte reductions of 90% for large fonts](https://www.w3.org/TR/PFE-evaluation/#analysis-cjk),
+
+- The dynamic subsetting severely impacted CDN **cache performance**
+- There was no subsetting of _design axes_, an unacceptable result with **variable fonts** becoming ubiquitous
+- Subsetting fonts with **complex interactions between glyphs** was challenging
+- Requiring an intelligent, dynamic server hindered widespread deployment
+- Very fine-grained subsetting [might be a privacy violation](https://www.w3.org/TR/2023/WD-IFT-20230530/#content-inference-from-character-set)
```
garretrieger, line 127: Another draw back is that it required a custom protocol (and http header) to communicate with the dynamic backend.

IFT-Explainer.md
```diff
@@ -0,0 +1,174 @@
...
+of patching a font to provide more data.
+However, instead of requiring custom patch generation for each user,
+the initial font has two new [_Patch Map_ tables](https://w3c.github.io/IFT/Overview.html#patch-map-dfn)
+which map each subset definition to **the link** of the relevant patch.
+Links are [stored in the font](https://w3c.github.io/IFT/Overview.html#uri-templates) 
+as RFC 6570 [URI Templates](https://www.rfc-editor.org/rfc/rfc6570).
+
+Thus, static hosting of files (base fonts, and patches) is easy,
+and cache performance is good
+because patches are shared between users.
```
garretrieger, line 142: Additionally this approach does not require any special http headers or a custom protocol to fetch patches. Requests are just normal GET requests. Which in turn means it's much easier to deploy and will work well with existing cdn infrastructure.

IFT-Explainer.md
```diff
@@ -0,0 +1,161 @@
...
+which the client applied to produce a new, enlarged subset font.
+
+It therefore required new server capabilities,
+in addition to client changes.
+
+Work on Range Request was discontinued because, despite
+[median byte reductions of 90% for large fonts](https://www.w3.org/TR/PFE-evaluation/#analysis-cjk),
+
+- The dynamic subsetting severely impacted CDN **cache performance**
+- There was no subsetting of _design axes_, an unacceptable result with **variable fonts** becoming ubiquitous
```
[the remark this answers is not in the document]
svgeesus: Ok so I should remove that bullet, then

IFT-Explainer.md
svgeesus: Added

IFT-Explainer.md
svgeesus: Fixed

IFT-Explainer.md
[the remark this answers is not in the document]
svgeesus: Removed

IFT-Explainer.md
```diff
@@ -0,0 +1,174 @@
...
+The [current specification](https://w3c.github.io/IFT/Overview.html) 
+draws on the Patch Subset concept
+of patching a font to provide more data.
+However, instead of requiring custom patch generation for each user,
+the initial font has two new [_Patch Map_ tables](https://w3c.github.io/IFT/Overview.html#patch-map-dfn)
+which map each subset definition to **the link** of the relevant patch.
+Links are [stored in the font](https://w3c.github.io/IFT/Overview.html#uri-templates) 
+as RFC 6570 [URI Templates](https://www.rfc-editor.org/rfc/rfc6570).
+
+Thus, static hosting of files (base fonts, and patches) is easy,
```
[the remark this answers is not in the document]
garretrieger: It's a pretty minor detail so I think it's fine to leave it out to keep things focused.

IFT-Explainer.md
```diff
@@ -0,0 +1,180 @@
...
+and cache performance is good
+because patches are shared between users.
+
+We incorporated a new idea to allow patches of glyph-only data to be **independent**.
+(This type of patch would have been too costly to compute dynamically, but they work well with this new framework of pre-computed patches.)
```
garretrieger, line 145: Suggest also adding that the independence allow for patches to be requested in parallel, which functionally is very similar to how unicode range webfont loading works. This helps make it clear what the benefits of independent patches are.

IFT-Explainer.md
```diff
@@ -0,0 +1,180 @@
...
+Thus, both **independent** (commutative) and **dependent** (non-commutative)
+patches are [supported](https://w3c.github.io/IFT/Overview.html#font-format-extensions).
+
+In addition, patches can add or extend design axes,
+to support variable fonts.
+
+It no longer requires any special HTTP headers,
+or a custom protocol to fetch patches.
+Requests are just normal HTTP GET requests,
+making it easier to deploy and working well
```
garretrieger, line 156: the wording "working well" seems slightly off in this sentence, maybe "compatible"?

garretrieger (approved):

IFT-Explainer.md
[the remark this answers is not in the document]
skef: Given that in the new system any dynamically generated patches still have pre-determined contents, it might take a few sentences to properly characterize what is the same and what changes. I would vote for leaving it out.

IFT-Explainer.md
```diff
@@ -0,0 +1,180 @@
...
+the initial font has two new [_Patch Map_ tables](https://w3c.github.io/IFT/Overview.html#patch-map-dfn)
+which map each subset definition to **the link** of the relevant patch.
+Links are [stored in the font](https://w3c.github.io/IFT/Overview.html#uri-templates) 
+as RFC 6570 [URI Templates](https://www.rfc-editor.org/rfc/rfc6570).
+
+Thus, static hosting of files (base fonts, and patches) is easy,
+and cache performance is good
+because patches are shared between users.
+
+We incorporated a new idea to allow patches of glyph-only data to be **independent**.
```
skef, line 144: "We also incorporated" would make this transition a bit less abrupt.

IFT-Explainer.md
svgeesus: Changed

IFT-Explainer.md
svgeesus: Added

PR #189, opened 2024-07-05
Updates for publication as WD on 2024 07 09

svgeesus:
- updates to local bibliography
- format as WD not ED
- sadly, some manual edits to the generated HTML to make it pass pubrules

Longer term this should be fixed by updating the bikeshed boilerplate files

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/189.html" title="Last updated on Jul 5, 2024, 5:49 PM UTC (c3946cd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/189/a8d734d...c3946cd.html" title="Last updated on Jul 5, 2024, 5:49 PM UTC (c3946cd)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -50,7 +50,7 @@ spec:fetch; type:dfn; for:/; text:response
     ],
     "status": "Internet Draft",
     "title": "Shared Brotli Compressed Data Format",
-    "date": "27 Jul 2021"
+    "date": "Sep 2022"
   },
 
   "open-type": {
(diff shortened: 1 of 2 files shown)


svgeesus:
See https://www.w3.org/TR/2024/WD-IFT-20240709/ which is not yet officially published (you won't find it from the /TR page and the history link is not yet updated)

garretrieger (approved):

skef:
I noticed reading through a previous draft that the instances of `*i*` in the paragraph starting "Assuming the subsetter" aren't getting modified in the HTML. I was probably thinking in terms of github markdown rather than whatever bikeshed uses. 

Anyway, this doesn't need to be fixed but my intention was that the "i"s would be in italic.

PR #190, opened 2024-07-10
Make Some Adjustments to Format 1 Patch Map

garretrieger:
I've been working on implementing a format 1 patch map parser and in the process found some issues and areas that were underspecified. This makes some changes to address those:
- Explicitly allow the feature map offset to be null.
- Change entry count from u32 -> u16. All entry index fields in the remainder of the table are at most u16 so no need for this to be u32.
- Rename "entry count" to "max entry index": an entry count of 0 is always invalid (entry 0 always exists). So changing this to "max entry index" avoids that. Additionally allows entry indices up to the max of u16 to be used.
- Specifically mention that entry 0 should be ignored (used to mark glyphs already in the font).
- Specifically mention entry indices greater than max entry index are invalid.
- Note that not all entry indices from 0 up to max entry index need to be encoded.
- Note that a single glyph may have multiple associated code points.

Note: this currently undoes the manual edits from https://github.com/w3c/IFT/pull/189, will need to fix that before it can be merged.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/190.html" title="Last updated on Jul 22, 2024, 7:56 PM UTC (4aae108)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/190/f0acde9...4aae108.html" title="Last updated on Jul 22, 2024, 7:56 PM UTC (4aae108)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -791,6 +791,9 @@ The algorithm:
 
 2.  For each unique <var>entry index</var> in [=Glyph Map/entryIndex=]:
 
+    *  If <var>entry index</var> is 0 then, this is a special entry used to mark glyphs which are already in the font. Skip this index
+        and do not build an entry for it.
+
     *  If the <var>entry index</var> is larger than [=Format 1 Patch Map/maxEntryIndex=] this table is invalid,
         return an error.
 
--- a/Overview.bs
+++ b/Overview.bs
@@ -619,8 +619,8 @@ the encoded bytes at the start of every iteration to pick up any changes made in
   </tr>
   <tr>
     <td>Offset32</td>
-    <td>featureMapOffset</td>
-    <td>Offset to a [=Feature Map=] sub table. Offset is from the start of this table.</td>
+    <td><dfn for="Format 1 Patch Map">featureMapOffset</dfn></td>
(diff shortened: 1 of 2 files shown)


skef (approved):
LGTM

svgeesus (approved):
This all seems reasonable to me.

svgeesus:
From https://github.com/w3c/IFT/pull/189#issue-2392927958

It would be useful to merge in the bibliographic updates. The date, and other changes to make it WD rather than ED can be skipped, now that publication has happened.

The boilerplate changes we should fix in the [bikeshed boilerplate repo](https://github.com/speced/bikeshed-boilerplate) instead of manually editing generated html.

garretrieger:
> From [#189 (comment)](https://github.com/w3c/IFT/pull/189#issue-2392927958)
> 
> It would be useful to merge in the bibliographic updates. The date, and other changes to make it WD rather than ED can be skipped, now that publication has happened.
> 
> The boilerplate changes we should fix in the [bikeshed boilerplate repo](https://github.com/speced/bikeshed-boilerplate) instead of manually editing generated html.

Created a PR for that: https://github.com/speced/bikeshed-boilerplate/pull/87

I'm going to manually edit on this PR to unblock this for now.

PR #191, opened 2024-07-22
Add additional restrictions to sparse bit set to avoid decoding values outside of unicode.

garretrieger:
1. Add a maximum height to each branch factor, this is the largest height needed to encode any 32 bit unsigned integer value for each branch factor.
2. In the decoding algorithm specify that values beyond the unicode maximum 0x10FFFF should be ignored.

The height maximum (above which the encoding will be rejected) is placed at 32 bits to allow room for any future changes to the unicode maximum, while allowing fonts using codepoints above 0x10FFFF to not be rejected by clients implementing this version of the standard.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/191.html" title="Last updated on Jul 22, 2024, 9:28 PM UTC (70f91e2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/191/a287147...70f91e2.html" title="Last updated on Jul 22, 2024, 9:28 PM UTC (70f91e2)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -1263,6 +1263,9 @@ each node has a fixed number of children that recursively sub-divides an interva
 branching factor <i>B</i> can store set membership for integers in the interval [0 to <i>B</i><sup><i>H</i></sup>-1] inclusive. The tree
 is encoded into an array of bytes for transport.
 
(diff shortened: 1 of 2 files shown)


skef (approved):
LGTM

PR #197, opened 2024-07-26
Additional adjustments to format 1 patch map

garretrieger:
Some further edits to format one to make it more in-line with the original IFTB proposal:
- Add a requirement for feature records to be sorted by tag, this was included in the original IFTB proposal.
- Fix the algorithm for producing entries from entry map records. Generate one output entry per entry map record, which matches the original IFTB proposal.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/197.html" title="Last updated on Jul 26, 2024, 8:36 PM UTC (fc9a55f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/197/6a10ae2...fc9a55f.html" title="Last updated on Jul 26, 2024, 8:36 PM UTC (fc9a55f)">Diff</a>


skef (approved):
I'm hazy on some of this stuff but what I'm not hazy on looks good.

Overview.bs
```diff
@@ -826,6 +826,9 @@ The algorithm:
     *  If the computed <var>mapped entry index</var> is larger than [=Format 1 Patch Map/maxEntryIndex=] this table is invalid,
         return an error.
 
+    *  If <var>mapped entry index</var> matches an entry index present in the [=Glyph Map=] this table is invalid,
```
skef, line 829: So, this is the equivalent to having the bin numbers for the feature patches be higher than those for the "base" glyph bins? Am I understanding this correctly?

garretrieger: That's right, I'm trying to cleanly separate the inputs and outputs of the feature mappings:
- input entry indices must come from the set of entries specified by glyph map.
- output entry indices must use any entry indices present in the glyph map.

I'm consider for simplicity that we may want to just use the approach you mentioned of requiring the feature output indices to be larger than the highest glyph map index (and the input indices to be less than).

skef: I was just confirming I understood the role of this change -- we can make things more specific later if we need to.

Issue #198, opened 2024-07-29
Should we require Erroring on Validation Requirements for Format 1 Patch Map?


garretrieger:
I've recently introduce some validation requirements to the format 1 patch map (#190 and #197). For example indices must be less than or equal to the reported maximum and in the feature map input indices may only come from the glyph map and output indices must not collide with anything in the glyph map.

Currently the specification text requires that when interpreting a format 1 table if data is found that violates these rules the entire interpretation operation fails with an error, which will in turn cause the extension operation to fail and prevent the client from further extending the font.

Checking these validation requirements would require the entirety of the format 1 table to be parsed up front. However, one of the intentions of the design of the format 1 glyph map is that it could be used in a lazy fashion. That is when locating entries which intersect the subset definition it could be done in a way where only the relevant parts of the format 1 table need to be accessed (eg. by using cmap + format 1 glyph map to locate the set of relevant entries). Thus requiring a full up front validation prevents the lazy approach from being taken by a client implementation and in some cases could significantly increase the computation cost in processing an IFT font with a very large glyph map.

Instead of making invalid data fail with an error I suggest we specify that any data which violates the validation rules is ignored. For example if a feature mapping references indices which are not in the glyph map then that particular mapping is ignored, but any other mappings which do not violate the rule may still be used. There's precedent for this in the other parts of font processing, for example shapers like harfbuzz do not error out on invalid data they just ignore it and work with whatever valid data remains.

skef:
> Instead of making invalid data fail with an error I suggest we specify that any data which violates the validation rules is ignored. 

How path dependent is this likely to be in practice? That is, will the same data tend to be ignored, or will what gets ignored change depending on which parameters (codepoints, etc.) are patched first? 

If the results will be somewhat path dependent it might be preferable to:

1. Add a validator to the reference suite of tools.
2. Declare the behavior of invalid encodings to be undefined. 

Or something along those lines.

garretrieger:
> > Instead of making invalid data fail with an error I suggest we specify that any data which violates the validation rules is ignored.
> 
> How path dependent is this likely to be in practice? That is, will the same data tend to be ignored, or will what gets ignored change depending on which parameters (codepoints, etc.) are patched first?
> 
> If the results will be somewhat path dependent it might be preferable to:
> 
> 1. Add a validator to the reference suite of tools.
> 2. Declare the behavior of invalid encodings to be undefined.
> 
> Or something along those lines.

[...]

PR #199, opened 2024-07-29
In format 1 ignore most invalid data instead of failing with an error.

garretrieger:
This allows a conformat client implementation to optimize format 1 entry intersection by only checking the relevant parts of the table. See #198 for more context.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/199.html" title="Last updated on Aug 27, 2024, 10:22 PM UTC (72967a4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/199/06f2625...72967a4.html" title="Last updated on Aug 27, 2024, 10:22 PM UTC (72967a4)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -618,8 +618,11 @@ the encoded bytes at the start of every iteration to pick up any changes made in
     <td>uint24</td>
     <td><dfn for="Format 1 Patch Map">glyphCount</dfn></td>
     <td>Number of glyphs that mappings are provided for.
-        Must match the numGlyphs field from [[open-type/maxp|maxp]] if present, or if numGlyphs is 65,535 then glyphCount
-        must be greater than or equal to 65,535.</td>
+        Must match the number of glyphs in the the font file.
+
+        Note: the number of glyphs in the font is encoded in the font file. At the time of writing, this value is listed in the
+        [[open-type/maxp|maxp]] table; however, future font format extensions may use alternate tables to encode the value for number of
+        glyphs.
   </tr>
   <tr>
     <td>Offset32</td>
--- a/Overview.bs
+++ b/Overview.bs
@@ -581,8 +581,6 @@ the encoded bytes at the start of every iteration to pick up any changes made in
 
 ### Patch Map Table: Format 1 ### {#patch-map-format-1}
 
-<!-- TODO: provide normative validation requirements for each encoding -->
-
 <dfn>Format 1 Patch Map</dfn> encoding:
 <table>
   <tr>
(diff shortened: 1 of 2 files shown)


svgeesus (approved):
These all seem like good changes and clarifications to me.

Not baking in the 64k limit is particularly nice

Overview.bs
```diff
@@ -609,9 +607,14 @@ the encoded bytes at the start of every iteration to pick up any changes made in
...
-    <td>uint32</td>
+    <td>uint16</td>
+    <td><dfn for="Format 1 Patch Map">maxGlyphMapEntryIndex</dfn></td>
+    <td>The largest [=Glyph Map=] entry index encoded in this table. Must be less than or equal to maxEntryIndex.</td>
+  </tr>
+  <tr>
+    <td>uint24</td>
     <td><dfn for="Format 1 Patch Map">glyphCount</dfn></td>
-    <td>Number of glyphs that mappings are provided for. Must match the numGlyphs field from [[open-type/maxp|maxp]].</td>
+    <td>Number of glyphs that mappings are provided for. Must match the numGlyphs field from either [[open-type/maxp|maxp]] or MAXP (if present). If both maxp and MAXP are present then, it must match the numGlyphs field in MAXP.</td>
```
skef, line 617: Are we referencing an ISO Working or Committee Draft?

skef: This may be a tricky question. Do W3C standards often reference "working" standards from other organizations? 

In any case I think it would be strange to reference that draft. I believe https://www.mpeg.org/wp-content/uploads/mpeg_meetings/146_Rennes/w23797.zip is not password-protected, but I'm not sure how stable the link will be.

garretrieger: This would be a good thing to discuss at our next call, I'm curious to get Vlad's thoughts.

I believe that we don't want to reference (in a normative way) anything that's a work in progress (eg. we have the same issue with the shared brotli spec). So maybe this is better put in as a note about possibly needing to integrate with beyond64k in the future (and we'll leave the field a u24 to support that).

skef: Yeah -- all of that stuff is (at least) months away from being an official part of the specification. Maybe years.

skef: (BTW if we just want a reference point for potentially having more than 64K glyphs, CFF2 already supports that. It's not very useful in practice because, e.g., hmtx doesn't, but it's a reference point.)

garretrieger: Reworked the text, now it only depends on maxp but allows the glyph count value to be greater than 65535 if maxp.numGlyphs is 65535. This will allow >16 bit glyph count to work without adding a dependency on MAXP. Once the boring expansion stuff is more official we can come back and update the requirements to also mention MAXP here.

garretrieger: Updated the wording here again based on the discussion from todays working group call. Please take another look.

skef (approved):
LGTM

PR #204, opened 2024-08-26
Clarify how to handle format 2 string entry id's.

garretrieger:
Updates the text to specify how to handle string ids for entries that don't set the optional entryIdStringLength field.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/204.html" title="Last updated on Aug 27, 2024, 12:01 AM UTC (dcbd8a5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/204/a80e684...dcbd8a5.html" title="Last updated on Aug 27, 2024, 12:01 AM UTC (dcbd8a5)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -1137,7 +1137,8 @@ The algorithm:
 1.  Check that the <var>patch map</var> has [=Format 2 Patch Map/format=] equal to 2 and is valid according to the requirements in
      [[#patch-map-format-2]] (requirements are marked with a "must"). If it is not return an error.
 
-2.  Initialize <var>last entry id</var> to 0, <var>current byte</var> to 0, and <var>current id string byte</var> to 0.
+2.  If the [=Format 2 Patch Map/entryIdStringData=] offset is 0 then initialize <var>last entry id</var> to 0. Otherwise initialize
+     it to an empty byte string. Set <var>current byte</var> to 0, and <var>current id string byte</var> to 0.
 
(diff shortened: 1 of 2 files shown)


Overview.bs
```diff
@@ -1137,7 +1137,8 @@ The algorithm:
 1.  Check that the <var>patch map</var> has [=Format 2 Patch Map/format=] equal to 2 and is valid according to the requirements in
      [[#patch-map-format-2]] (requirements are marked with a "must"). If it is not return an error.
 
-2.  Initialize <var>last entry id</var> to 0, <var>current byte</var> to 0, and <var>current id string byte</var> to 0.
+2.  If the [=Format 2 Patch Map/entryIdStringData=] offset is 0 then initialize <var>last entry id</var> to 0. Otherwise initialize
```
skef, line 1140: I haven't been tracking our terminology closely enough: How are we using "array" in this document generally? When I think about vectors, strings, and arrays, the former are variable length while the latter isn't, but I guess that's a C++-centric way of looking at things. (So to me an "empty byte array" seems like something that would only be appropriate in certain degenerate cases.)

skef (approved):
LGTM

Overview.bs
garretrieger: Looking at other text related to this field we call it a byte string, so changed the wording here to "empty byte string".

Issue #214, opened 2024-09-24
Remove full file brotli patch mechanism


rsheeter:
Per f2f at TPAC

rsheeter:
and rename Per Table Brotli to better reflect what over how, be a better partner to Glyph Keyed

PR #217, opened 2024-09-30
Remove the pure brotli patch type, rename remaining brotli patch type to "Table Keyed"

garretrieger:
As discussed at the TPAC working group meeting this drops the "brotli" patch type because it is not expected to see wide use due to incompatibility with WOFF2 and mixed patch type encodings.

This leaves two remaining patch types:
- Table Keyed (renamed from the previous "per table brotli"): operates at the table level.
- Glyph Keyed: operates at the glyph level.

"Per table brotli" was renamed to "Table Keyed" to keep the naming of the patch formats consistent. The mention of brotli was dropped from the name since both glyph keyed and table keyed patches use brotli internally.

Fixes: #214

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/217.html" title="Last updated on Sep 30, 2024, 10:10 PM UTC (0eab2f7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/217/14ee493...0eab2f7.html" title="Last updated on Sep 30, 2024, 10:10 PM UTC (0eab2f7)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -579,7 +579,7 @@ Incremental Font Transfer and WOFF2 {#ift-and-woff2}
 is taken. If an incremental font will be encoded by WOFF2 for transfer:
 
 1.  If the WOFF2 encoding will include a transformed glyf and loca table ([[WOFF2#glyf_table_format]]) then, the incremental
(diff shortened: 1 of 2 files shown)


skef (approved):
Looks good!

PR #222, opened 2024-10-21
Add mechanism for handling UVS sequences (and other similar cases)

garretrieger:
Draft of a possible solution for allowing IFT to efficiently handled things like Unicode variation selectors. These require being able to specify patches which are only loaded if both a modifier codepoint and one or more base codepoints are simultaneously present. The existing patch mapping mechanisms don't currently provide any way to do this.

This modifies the logical entry structure to allow it to have one or more subset definitions (previously was always one), where each subset def must match the clients content in order for the entry to be selected. Then the format 2 copy indices mechanism is extended to provide a way to encode entries with multiple subset definitions.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/222.html" title="Last updated on Nov 4, 2024, 4:30 PM UTC (4cd5d57)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/222/45dd96d...4cd5d57.html" title="Last updated on Nov 4, 2024, 4:30 PM UTC (4cd5d57)">Diff</a>


garretrieger:
Per #223 we will need a solution for readiness detection that works with this approach before we could incorporate this new mechanism.

skef:
I've read through this but I guess the more basic comprehension problems I have with the specification mean that I can't effectively review. I don't fully understand the new patch format/algorithms.

garretrieger:
> I've read through this but I guess the more basic comprehension problems I have with the specification mean that I can't effectively review. I don't fully understand the new patch format/algorithms.

You can hold off on reviewing this for now, we'll need to figure out the readiness stuff first and there's a chance this may just  abandoned based on that.

As for improving comprehension I don't have it ready yet, but I am working on adding full examples of the extension algorithm execution to demonstrate how everything works together. This PR does add some examples of how the intersection step works which you may find that helpful for illustrating the high level outcome of this change (particularly the last two rows in the table show the new functionality).

skef:
I've been thinking more about "readiness" and this proposal and trying to nail down what, if anything, the concern would be.

Suppose we set aside glyph-keyed patches for a moment and just talk about table-keyed patches (in the patch-subset style). And we imagine that there are ligature-specific table-keyed patches. In that case, if one of those is not applied when the client is temporarily rendering content, the "cost" is temporarily not seeing the ligature and just seeing the "base glyphs". I think that would be fine, as it is in line with the general phenomenon of temporary rendering.

What I hadn't put together until now is the implications of glyph-keyed patches, either combined with table-keyed patches or used on their own. (You're probably only going to ever use this functionality in a glyph-keyed patch, and I knew that, I just hadn't realized the following.) If you have the shaping rules for a ligature (or UVS, or whatever else falls under this category) but not the glyph outline, you'll render a blank. Maybe this is what @garretrieger has been worried about the whole time, and we just didn't get into enough specifics to make it clear to me.

I guess I've made it obvious that I strongly believe readiness should be secondary to patch functionality and flexibility, so I'm bothered by the idea of leaving this out just for the sake of readiness. However, I also see the problem. So with both of those things in mind, I think the next question should be: what can we say about readiness in light of this proposed mechanism?

Thinking about this, I think I'm going to go back on what I've been saying in the readiness PR. A client probably should be "shaping" some substrings -- even individual codepoints -- independently. The question is whether we can provide specific guidance.

If you have a patch that adds support for codepoint *x* (or *may* add support for it), we basically advise (or imply) that the client shouldn't attempt to use that codepoint with the font -- it should fall back. We provide a procedure for making that determination. 

If we have this new flavor of patch, and one of them isn't loaded into the current font, then I think the advice should be something akin to:

> If the per-codepoint test succeeds, but there is a patch identified as required that relates to the codepoint, the patch corresponds to inter-codepoint dependencies. In that case the codepoint can be used with the font for temporary shaping, but the shaping unit must not contain codepoints from more than one group referenced by the patch. Therefore, if there is more than one such codepoint in the original shaping unit it must be broken into subunits that are shaped separately.

If that's correct at a high level, then the trick is turning this into real language instead of the hand-wavy way I've phrased it.

garretrieger:
> I've been thinking more about "readiness" and this proposal and trying to nail down what, if anything, the concern would be.
> 
> Suppose we set aside glyph-keyed patches for a moment and just talk about table-keyed patches (in the patch-subset style). And we imagine that there are ligature-specific table-keyed patches. In that case, if one of those is not applied when the client is temporarily rendering content, the "cost" is temporarily not seeing the ligature and just seeing the "base glyphs". I think that would be fine, as it is in line with the general phenomenon of temporary rendering.
> 
> What I hadn't put together until now is the implications of glyph-keyed patches, either combined with table-keyed patches or used on their own. (You're probably only going to ever use this functionality in a glyph-keyed patch, and I knew that, I just hadn't realized the following.) If you have the shaping rules for a ligature (or UVS, or whatever else falls under this category) but not the glyph outline, you'll render a blank. Maybe this is what @garretrieger has been worried about the whole time, and we just didn't get into enough specifics to make it clear to me.
> 
> I guess I've made it obvious that I strongly believe readiness should be secondary to patch functionality and flexibility, so I'm bothered by the idea of leaving this out just for the sake of readiness. However, I also see the problem. So with both of those things in mind, I think the next question should be: what can we say about readiness in light of this proposed mechanism?
> 
> Thinking about this, I think I'm going to go back on what I've been saying in the readiness PR. A client probably should be "shaping" some substrings -- even individual codepoints -- independently. The question is whether we can provide specific guidance.
> 
> If you have a patch that adds support for codepoint _x_ (or _may_ add support for it), we basically advise (or imply) that the client shouldn't attempt to use that codepoint with the font -- it should fall back. We provide a procedure for making that determination.
> 
> If we have this new flavor of patch, and one of them isn't loaded into the current font, then I think the advice should be something akin to:
> 
> > If the per-codepoint test succeeds, but there is a patch identified as required that relates to the codepoint, the patch corresponds to inter-codepoint dependencies. In that case the codepoint can be used with the font for temporary shaping, but the shaping unit must not contain codepoints from more than one group referenced by the patch. Therefore, if there is more than one such codepoint in the original shaping unit it must be broken into subunits that are shaped separately.
> 
> If that's correct at a high level, then the trick is turning this into real language instead of the hand-wavy way I've phrased it.

I spent some time thinking about this and came up with a reasonably simple algorithm that uses the idea of locating sub spans within the shaping unit which are supported by the font. Sent that out as a new PR: https://github.com/w3c/IFT/pull/227.

svgeesus (approved):
This seems significantly clearer, to me.

skef (approved):

Issue #223, opened 2024-10-21
Detecting codepoint readiness


garretrieger:
This is a sub issue of #201. In general the client will need to be able to detect at any point which codepoints/features/design space combintations in an IFT font are ready to use so that the client can apply font fallback strategies if the unit is not ready yet. I've thought about some different approaches to how this could be done and came up with the following approach:

- Given a span of text that the client is looking to use an IFT font for it can assign each code point in that span the specific set of features being used and the specific point in design space that the code point will be rendered with. Note: here we're looking at the ordered set of code points, so the same codepoint may appear multiple times with different feature and design space configurations.
-  In patch mappings individual code points match patch entries completely independent of other code points, since we use set intersection when checking entries.
- Using the above two points the client can check for each code point in the span if there are any unapplied entries which would match that code point + feature + design space combination.
- If one more more unapplied entries matches then the client will consider that code point to not be ready and apply whatever usual fallback strategies it would normally use. Otherwise the code point can be rendered with the IFT font.
- I believe this should be easy to incorporate into browser's existing fallback handling. For example from what I know of chromes handling of unicode range fallback this would fit in pretty naturally.

We can incorporate this into the spec by adding an additional section which provides an algorithm clients can use to determine readiness.

This appears to me to handle the general case of checking readiness both for the initial font and at any point during the augmentation process for a font. Unfortunately however, it will no longer work if we adopt the new mechanism proposed in #222. This is because that PR would make patch selection no longer independent per codepoint. For example consider the case where data is already loaded for codepoint a and it's considered ready but then later on a UVS selector gets added to the text used with codepoint a where the variant glyph for a isn't yet loaded. To be correct we wouldn't want to consider (UVS + a) ready, but that requires the client considering combinations of codepoints in the input text span when checking readiness. Which in turn requires understanding how codepoints interact inside of the font. Codepoints don't always need to be adjacent to interact. Thus it isn't always possible without having a fully loaded the relevant font data, since the patch mapping doesn't capture how the codepoints interact only that they do in some way.

As a result, I think we may want to consider not adopting the mechanism proposed in #222 because it will make determining readiness in the browser far more difficult and complicated. We are unfortunately forced to make a tradeoff here and reduce our ability to optimally encode certain cases like UVS selectors, but ultimately I think it's more important that the client have a clean and easy way to detect readiness on a per codepoint basis (to align with existing fallback handling).

[...]

Issue #201, opened 2024-08-13
Considerations for "desiccated" initial font files


skef:
Apologies if this is handled somewhere in the spec. I don't remember it being discussed explicitly but maybe it's all implicit.

I know we've discussed the possibility of effectively removing a round-trip by loading the initial font file in CSS, so that it functions analogously to the unicode mapping for unicode-range. I believe this could be done with a Base64-encoded data URL, and maybe it could be done in other ways. 

Suppose (because you're paying a penalty for the Base64) you want to minimize the size of the initial file, downloading most of the "universal" data via an initial patch. So let's assume we're doing some combination of per-table patches and glyph-keyed patches (one or the other or both). 

- What is the minimum initial table set assuming per-table patches? My previous understanding is that it would be cmap, IFT, and IFTX (when present) but maybe things have expanded since then? Are there certain tables that should be required in order to solve chicken-and-egg problems on the client side -- perhaps "head" or "post"? 
- For glyph-keyed patches the normal presumption is that the glyf/loca/GVAR or CFF/CFF2 tables will be present. Maybe this is fine, they should compress pretty well when "empty". If we want that not to be the case, though, and to download those tables in the initial per-table patch, can we have some semantic for doing that without an extra round-trip? (That is, to start downloading the initial per-table patch and the glyph-keyed patches at the same time, but guarantee that the per-table patch is applied first.)
- Should we add language to the effect of "an initial font not yet checked for some set of code points cannot be judged for completeness or OpenType/OFF spec compliance."?

garretrieger:
> * What is the minimum initial table set assuming per-table patches? My previous understanding is that it would be cmap, IFT, and IFTX (when present) but maybe things have expanded since then? Are there certain tables that should be required in order to solve chicken-and-egg problems on the client side -- perhaps "head" or "post"?

Format 1 patch maps have a dependency on cmap and maxp. Format 2 patch maps do not require any other tables. As you stated glyph keyed patches will additionally depend on a loca + glyf/CFF/CFF2 table being present. I don't believe head or post would be needed so long as the IFT processing happens before the client attempts to use the font in anyway.

[...]

PR #224, opened 2024-10-25
Add a procedure to check what content an IFT font can render.

garretrieger:
This PR defines how a client can form target subset definitions and provides an optional procedure the client can use to determine what parts of content to be rendered can be supported by an IFT font in a particular state.

This adds the changes suggested in #223, and provides a concrete way to assess the state of an initial IFT font for #201.

This is currently written with respect to the current state of the specification and would need to be updated if we decide to adopt the changes in #222.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/224.html" title="Last updated on Oct 29, 2024, 8:19 PM UTC (7b72bca)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/224/11aba9e...7b72bca.html" title="Last updated on Oct 29, 2024, 8:19 PM UTC (7b72bca)">Diff</a>


Overview.bs
```diff
@@ -527,15 +527,68 @@ The algorithm:
...
+
+* For each shaping unit there are two checks:
+
+    * First, if for any code point in the shaping unit there is not a [[open-type/cmap|cmap]] entry for it, or the entry maps to glyph
+        0 then the incremental font does not fully support rendering the shaping unit.
+
+    * Second, compute the corresponding [=font subset definition=] and execute the [$Extend an Incremental Font Subset$] algorithm,
+        stopping at step 6. If the entry list is not empty then the incremental font does not fully support rendering the shaping unit.
+
+In most cases the client may also wish to know what the font can render at a more granular level than a shaping unit. The following
```
skef, line 572: On tone: I'd prefer making this and the below less, well, emphatic, I guess. Why "most" here instead of "some"? And why "In some rare cases" below? The subject here is a procedure that can be used for temporary rendering when the font is in an complete state relative to a full shaping unit. Let's just lay it out and be clear that even if all the codepoints are present in the narrow sense the overall result may be different and will have to be re-rendered once the font is patched.

On substance: As the implementer of a client, is what I want to know really "what the font can render at a more granular level than a shaping unit"? It seems more like what I want to know is *how* to (temporarily) render content with the font at its current patch level, to the extent I can use the font. So: we've described the codepoint-by-codepoint check, and that tells us what definitely *can't* be rendered (without risking a "blank"). With that information, the process is presumably something like: 

 * Break the shaping-unit up into sub-units at either side of each sequence of unsupported codepoints (if any).
 * Render each sub-unit separately, with unsupported codepoint sequences rendered using the fallback font.
 * Piece the sub-units back together.

And we could note (again) that if there are no unsupported codepoints, the shaping unit can be rendered together but the result may not be the same as if the patched identified by the extension algorithm were loaded.

Maybe this is all implicit, but I feel like we should at least gesture to something along these lines because talking about treating codepoints individually might lead some readers to think that they should be rendered individually, which is not a good approach.

garretrieger: > On tone: I'd prefer making this and the below less, well, emphatic, I guess. Why "most" here instead of "some"? And why "In some rare cases" below? The subject here is a procedure that can be used for temporary rendering when the font is in an complete state relative to a full shaping unit. Let's just lay it out and be clear that even if all the codepoints are present in the narrow sense the overall result may be different and will have to be re-rendered once the font is patched.

Sounds good, I'll adjust the wording.

> 
> On substance: As the implementer of a client, is what I want to know really "what the font can render at a more granular level than a shaping unit"? It seems more like what I want to know is _how_ to (temporarily) render content with the font at its current patch level, to the extent I can use the font. So: we've described the codepoint-by-codepoint check, and that tells us what definitely _can't_ be rendered (without risking a "blank"). With that information, the process is presumably something like:
> 
> * Break the shaping-unit up into sub-units at either side of each sequence of unsupported codepoints (if any).
> * Render each sub-unit separately, with unsupported codepoint sequences rendered using the fallback font.
> * Piece the sub-units back together.

Adding this would essentially just be telling clients how to implement font fallback rendering. Which I feel like is out of scope for this spec. Clients should already have an existing strategy for fallback rendering since the problem exists with regular fonts too. All we need to do in this specification is tell them how to detect where they need to fallback. Normally a client could rely on cmap (or other mechanisms) to do this, but with an IFT font cmap doesn't always accurately represent the capabilities so this more complicated check is needed.

It sounds like I should make this more explicit though and mention that this procedure is meant to slot into existing fallback algorithms as an extension to the usual cmap (or whatever the client normal does) mechanism for determining where to fallback.

garretrieger: Also I think I'll add a sentence to call out that clients should not rely on cmap being accurate for IFT fonts.

skef: I agree that we probably shouldn't be requiring a specific procedure. On the other hand, there's a general goal or principle of shaping "the largest possible subunits" together that I think is objectively correct compared with the alternatives and that would be good to express somehow. Generally speaking the more you break things apart the more likely you are to get differences, so clients should only break things apart to the extent they need to to implement fallback. Or at least, that's how I've come to think about it.

skef (approved):
I'd like to see something discouraging rendering codepoint-by-codepoint as opposed to checking codepoint-by-codepoint but I won't hold up this PR for that.

PR #226, opened 2024-10-29
Add note about using unicode-range + IFT.

### LINKED ISSUE #210: Advice for `unicode-range`
In [Opt-In Mechanism](https://w3c.github.io/IFT/Overview.html#opt-in) a CSS snippet is given to request an IFT font.

There is no mention of `unicode-range` and I can imagine content authors coming to one of three conclusions here:

1. This is a subset font, so specify the  `unicode-range` or the subset.
2. This is an IFT font, so specify the  `unicode-range` for the fully expanded font
3. This is an IFT font, so could contain anything, so use the [initial value, `U+0-10FFFF`](https://drafts.csswg.org/css-fonts-4/#unicode-range-desc)

I think option 1 is clearly wrong, because CSS will never use a font outside the declared `unicode-range`. So we should explicitly say not to do that.

I think option 2 is correct, but the user may not have convenient access to the fully expanded font to run tooling that deduces the `unicode-range`.

garretrieger:
Fixes #210


svgeesus (approved):
LGTM

PR #227, opened 2024-10-29
Add a more general readiness check for finding which parts within a shaping unit are supported.

garretrieger:
Insted of checking code point by code point, greedily accumulate spans of text which do not have any matching unnapplied patches. Each span can be safely rendered in isolation, however substitution rules between spans will not work until font extension is complete for the shaping unit.

This works on the following assumptions:
- By encoder requirements, if we have a span of text with an associated subset definition and there are no unapplied patches matching that definition then that span of text can be rendered correctly in isolation.
- Then while a shaping unit as a whole is not ready we can still find spans within that unit which are ready.

Given that, we can then use a straightforward greedy approach where we incrementally increase a span until it fails the intersection check. This then repeats to form as many spans as possible within the shaping unit. Code points which are not currently supported at all will not be included in any of the returned spans, and those should be rendered with a fallback.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/227.html" title="Last updated on Oct 30, 2024, 11:04 PM UTC (8cb7b66)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/227/2ad687e...8cb7b66.html" title="Last updated on Oct 30, 2024, 11:04 PM UTC (8cb7b66)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -584,23 +584,21 @@ incremental font:
 # supported by and can be safely rendered with ift_font.
 #
 # shaping_unit is an array where each item has a code point, associated list of
-# features, and the design space point that code point is being rendered with.
+# layout features, and the design space point that code point is being rendered with.
 def supported_spans(shaping_unit, ift_font):
-  current_start = None
-  current_end = None
-  current_subset_def = None
-
+  current_start, current_end, current_subset_def = None
   supported_spans = []
 
-  while i < len(shaping_unit):
+  i = 0
+  while i < shaping_unit.length():
     if current_subset_def is None:
       current_subset_def = SubsetDefinition()
       current_start = i
 
     current_end = i
-    current_subset_def.add(shapping_unit_text[i].codepoint,
-                           shapping_unit_text[i].features,
-                           shapping_unit_text[i].design_space_point)
+    current_subset_def.add(shaping_unit.codepoint_at(i),
+                           shaping_unit.features_at(i),
+                           shaping_unit.design_space_point_at(i))
 
     if not has_unapplied_patches(ift_font, current_subset_def):
       i += 1
@@ -613,9 +611,7 @@ def supported_spans(shaping_unit, ift_font):
     else:
       i += 1
 
-    current_subset_def = None
-    current_start = None
-    current_end = None
+    current_start, current_end, current_subset_def = None
 
   return supported_spans
 
--- a/Overview.bs
+++ b/Overview.bs
@@ -575,29 +575,69 @@ parts of some content an incremental font can render:
 
 * Any shaping units that passed both checks can be rendered in their entirety with the font.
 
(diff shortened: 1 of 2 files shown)


Overview.bs
```diff
@@ -575,29 +575,69 @@ parts of some content an incremental font can render:
...
+# shaping_unit is an array where each item has a code point, associated list of
+# features, and the design space point that code point is being rendered with.
+def supported_spans(shaping_unit, ift_font):
+  current_start = None
+  current_end = None
+  current_subset_def = None
+
+  supported_spans = []
+
+  while i < len(shaping_unit):
```
skef, line 595: I'm not sure of current pseudocode conventions. Is it overkill to initialize `i` to 0?

Overview.bs
```diff
@@ -575,29 +575,69 @@ parts of some content an incremental font can render:
...
+    if current_subset_def is None:
+      current_subset_def = SubsetDefinition()
+      current_start = i
+
+    current_end = i
+    current_subset_def.add(shapping_unit_text[i].codepoint,
```
skef, line 601: Should be `shaping_unit_text` or maybe just `shaping_unit` (Learned something from Urban Dictionary today ...)

This is picky but I'm wondering about maybe using a different abstraction like:

```
     shaping_unit.codepoint_at(i),
     shaping_unit.features_at(i),
     shaping_unit.design_space_point_at(i)
```

I'm just thinking that the latter two parameters are normally going to be represented by spans so a programmatic abstraction could ward off naive interpretations.

Overview.bs
```diff
@@ -575,29 +575,69 @@ parts of some content an incremental font can render:
...
+incremental font:
+
+<pre highlight="python">
+# Returns a list of spans, [start, end] inclusive, within shaping_unit that are
+# supported by and can be safely rendered with ift_font.
+#
+# shaping_unit is an array where each item has a code point, associated list of
+# features, and the design space point that code point is being rendered with.
+def supported_spans(shaping_unit, ift_font):
+  current_start = None
```
skef, line 589: This and the below are pseudo-code cases where I like chaining such assignments on one line, I guess based on a lines-are-emphasis way of thinking. But YMMV.

Overview.bs
garretrieger: I think it makes sense to have, added.

Overview.bs
garretrieger: Fixed and made everything consistent, settled on "shaping_unit".

garretrieger: Also changed to .codepoint_at(i), .feature_at(i), and .design_space_point_at(i)

Overview.bs
garretrieger: Sounds good, fixed

skef (approved):
I tweaked the chained assignments.

This all looks great!

Issue #216, opened 2024-09-24
Add concrete examples to the application of a patch section


rsheeter:
Per f2f TPAC discussion, attempt to make the interaction of encoding chocies and application of a patch clearer by adding concrete example(s)

vlevantovsky:
Also, based on the input from outside reviewers - providing examples of full and partial patch invalidations, along with more descriptive text, would be useful.

PR #237, opened 2024-11-14
Add appendix B which contains an example of the extension algorithm execution.

garretrieger:
The example shows howed a mixed (table + glyph keyed) IFT font would typically be processed by a client.

For #216

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/237.html" title="Last updated on Dec 3, 2024, 5:01 PM UTC (b918bb4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/237/457cd10...b918bb4.html" title="Last updated on Dec 3, 2024, 5:01 PM UTC (b918bb4)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -2884,7 +2884,7 @@ Inputs:
 
 Iteration 1:
 
-*  Steps 1 through 6 - applying the [$Check entry intersection$] check to all entries, the following patch entries intersect:
+*  Steps 1 through 6 - applying the [$Check entry intersection$] check to all entries in the initial font, the following patch entries intersect:
     * //foo.bar/01.tk
     * //foo.bar/02.tk
     * //foo.bar/04.tk
@@ -2928,7 +2928,8 @@ Iteration 1:
 
 Iteration 2:
 
-*  Steps 2 through 6 - applying the [$Check entry intersection$] check to all entries, the following patch entries intersect:
+*  Steps 2 through 6 - applying the [$Check entry intersection$] check to all entries in the updated font from the previous iteration, the following patch
+    entries intersect:
     * //foo.bar/01.gk
     * //foo.bar/04.gk
 
@@ -2943,7 +2944,8 @@ Iteration 2:
 
 Iteration 3:
 
-*  Steps 2 through 6 - applying the [$Check entry intersection$] check to all entries, the following patch entries intersect:
+*  Steps 2 through 6 - applying the [$Check entry intersection$] check to all entries in the updated font from the previous iteration, the following patch
+    entries intersect:
     * //foo.bar/04.gk
 
 *  Step 7 - one entry remains, it is selected.
--- a/Overview.bs
+++ b/Overview.bs
@@ -500,6 +500,10 @@ be needed in later iterations.
 Note: if the only remaining intersecting entries are no invalidation entries the intersection check in step 5 does not need to be repeated on each
 iteration since no invalidation patches won't change the list of available patches upon application (other then to remove the just applied patch).
 
+<div class="example">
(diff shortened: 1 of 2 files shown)


Overview.bs
```diff
@@ -2671,3 +2675,227 @@ to be used by default in most shaper implementations. This list was assembled fr
...
+    <td>//foo.bar/05.gk</td>
+    <td>
+      3, Glyph Keyed
+    </td>
+  </tr>
+</table>
+
+<b>Optimized Extension Example</b>
+
+Note: this example execution is described as it would be implemented in an optimized client which aims to reduce the number of round trips needing to be
```
skef, line 2815: I think one aspect of this description needs updating in light of #238 (as expected) and the second isn't accurate relative to the example. This is will now be following the selection criteria rather than "making choices". And I don't think it's doing any speculative loading -- it's choosing 3 as it should. Or maybe that's what was meant. 

I think going forward we can talk about the algorithm now that it's nailed down, and occasionally talk about speculative loading. For the latter I want to get away from the term "preload" because we're using that for adding patches to a file on the server side.

garretrieger: I've updated the later step that talks about invalidating patch selection to reflect the changes in #238. Also as suggested changed preload to speculative.

For this particular note the speculative loading I'm talking about is for the glyph keyed patches (so not related to invalidating patch selection), I added some text to clarify that.

skef: > For this particular note the speculative loading I'm talking about is for the glyph keyed patches (so not related to invalidating patch selection), I added some text to clarify that.

The set requires 'f' and 'P' and we load 01.gk and 04.gk, so I don't see any speculative loading for those.

garretrieger: These are speculative in the sense that we have to apply the table keyed patch first and we don't know for sure exactly what it will do. Although unlikely it could modify the entry list for the glyph keyed patches and remove 01.gk and 04.gk entries making the loads for those two unnecessary. Maybe speculative isn't quite the right word here, how about preemptive?

skef: OK, hmm. From the text further down:

> The table keyed patch //foo.bar/04.tk is partially invalidating which means the two glyph keyed patches will be remain valid after the application of //foo.bar/04.tk. Thus the client can expect that the two glyph keyed patches will be required in future iterations and preemptively begin the loads for those two patches.

What's ambiguous about the situation such that either "speculative" or "preemptive" really applies? If 04.tk *could* make the other two patches redundant, it would be one thing, but it can't. 

I suppose loading patches in parallel can be seen as an "optimization" but in practice it will kill performance not to, so I don't think we should present it as "optional" except in the loosest sense.

Why is there any linkage between the two patch groups in the algorithm in the first place? If 01.gk arrives before 04.tk, can't you (in terms of correctness) apply it first?

garretrieger: Yes, the invalidation modes guarantees that the glyph keyed patches in this case are always safe to apply. However, the invalidation rules don't prevent a table keyed patch from rewriting the glyph keyed map to point to a different set of URIs that still use the same compatibility base (eg. how we handle variable axis extension in the prototype). So there's no harm in applying the glyph keyed patches early, but it's possible (though unlikely) they may become unnecessary if they are superseded by new patches. Definitely agree that saving the round trip is always worth it even in the odd case where you grab patches you don't end up needing. I'll reword this a bit to tone down the "speculative" nature of this behaviour and just present it as what a good implementation should do.

garretrieger: Ok, updated the text. What do you think?

Issue #235, opened 2024-11-08
Potential invalidating patch criterion for extension algorithm


skef:
I suspect that the lack of a specific criterion for choosing the next patch during extension will actually interfere with generality. The more information an encoder has about what the client will do, the more clever and "general" it can be about how it encodes. As I've indicated in #232 I think the considerations for glyph-keyed patches will be distinct from those for invalidating patches so I'm only speculating about the latter here. I'm also going to skip the variable subspace problem because I'd need to think more about how to add that in to this framework.

My thoughts about this have evolved primarily from thinking about "one-round-trip encodings" -- encodings such that each patch map has a patch that loads all remaining parameters, and then other patches that load various subsets of those parameters. How do you organize the overall system sensibly to allow for that?

Let's think about two per-patch measures:

 * StuffYouNeed: How many of the codepoints and features needed are loaded by the patch. For now I'm happy to think about this as the simple sum of the two, meaning that the client can easily calculate this using some set operations. (This is setting aside the nascent UVS-motivated mechanism.)
 * TotalStuff: "How much" is loaded by the patch. This is impractical for the client to measure as things stand.

If you *could* measure TotalStuff, it seems like a good criterion would be:

  * Choose the patches that maximize StuffYouNeed
  * Among those, choose the patch that minimizes TotalStuff, choosing arbitrarily among ties.

This is a very simple strategy and seems like a good starting point for discussion.

Now, since TotalStuff is hard to calculate one approach would be to add a numeric field representing that number to the patch map. I think we've discussed something along these lines in the past with the expected distaste. What does it represent? How would the encoder calculate it? Do we really need to pay the cost for it in the patch table? What about patches listed multiple times? etc. I share these negative thoughts.

Happily, we don't really care the numeric value, it's just there to break ties relative to StuffYouNeed. And *I think* we're not currently using the patch map *order* for anything. (It clearly doesn't factor into the extension algorithm, but there might be order-based size optimizations I'm not remembering. If there are, think in terms of an added numeric field to represent the TotalStuff *order*, and unpacking the patch maps into those orders before analysis.) So the encoder could order the patches in the maps by "increasing TotalStuff", and the second part of the updated criterion would be:

 * Among those, choose the first patch.

So in practice the algorithm would just maintain the current patch maximizing StuffYouNeed as it goes through the list.

[...]

PR #238, opened 2024-11-15
Introduce a specific criteria for selecting one of many candidate invalidating patches during extension.

garretrieger:
- For a set of patches who have the same intersection with the target subset definition and where that intersection is not a strict subset of any other intersections we know that exactly one of those patches will always be needed regardless of any other patch selections.
- From there pick the one that has the smallest transfer size.
- Here we use the ordering of a patch in the patch map as a signal for the relative transfer size of that patch.
- Lastly, update encoding guidance to recommend encoders order map entries by transfer size ascending.

For #235

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/238.html" title="Last updated on Nov 15, 2024, 9:25 PM UTC (bcf907b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/238/5068d03...bcf907b.html" title="Last updated on Nov 15, 2024, 9:25 PM UTC (bcf907b)">Diff</a>


svgeesus (commented):
This seems to capture what is in https://github.com/w3c/IFT/issues/235 but commenting rather than approving, defer to @skef

skef (approved):
Very good -- I was wondering if we should say more about the "an entry" in step 2, but I like the way you've handled the question in the Note. Given how we expect patches to be arranged the choice among the independent (same-intersection) patch groups would be basically arbitrary, as each one will have its own round trip. Suggesting loading the most StuffYouNeed first but not requiring that is reasonable.

Issue #239, opened 2024-11-19
Alternative high-level breakdown of patch extension algorithm


skef:
We've already discussed changing the current description so that instead of creating a single list of patches, we create lists by invalidation status: full, partial, or no. When we discussed this the idea was to process them in that order. We've also discussed using the order of entries in the map for tie breaking, so we'll probably want to say that when the patches are broken out into these lists the ordering must be preserved.

In addition to that, I propose that we change the description to say that the list of full invalidation patches (if any) must be processed *first*. Then, once there are no remaining intersecting full invalidation patches, the remaining two lists can be processed *concurrently*. This makes the nature of partial invalidation patches clear in the algorithm and is better relative to the concerns of [this thread](https://github.com/w3c/IFT/pull/237#discussion_r1844900800).

garretrieger:
I experimented with a few different ways of approaching this and settled on #244. The approach I ended up going with is adding an explicit step to the extension algorithm that instructs the client to start loads for any entries which won't be invalidated by the current selection. The reason I did it this way is:

- Patch application by it's nature is will always be done serially so I want to keep the extension algorithm with one patch application per iteration as it's currently written.
- However, patch loads are free to run concurrently so that's what I addressed vs making the high level processing concurrent as suggested in this issue.
- As discussed in [this comment](https://github.com/w3c/IFT/issues/240#issuecomment-2489640938) we can't always blindly load all no invalidation patches when processing a partial invalidation patch. Instead the correct thing to do is to start loads for all things that you can see won't be invalidated by the current selection.
- Performance concerns in #232 are resolved by instructing the client to skip redundant intersection checks.

Overall I believe this addresses the concerns raised here and related issues:
- We now have clear and simple guidance for what should be concurrently fetched, providing clarity for encoder implementations so they will be able to predict what will actually happen in an optimized client implementation.
- This guidance respects the invalidation behaviour as currently defined.
- Improved clarity for how processing prioritizes invalidation types by grouping entries by invalidation type prior to selection.

garretrieger:
Should be addressed by the #244, but feel free to reopen if more is needed.

Issue #233, opened 2024-11-08
Grouping of patches during extension


skef:
Related to #232 

The extension algorithm says:

> For each of tables 'IFT ' and 'IFTX' (if present): convert the table into a list of entries by invoking Interpret Format 1 Patch Map or Interpret Format 2 Patch Map. Concatenate the returned entry lists into a single list, entry list.

And then later:

> Pick one entry from entry list with the following procedure:

> * If entry list contains one or more patch map entries which have a patch format that is Full Invalidation then, select exactly one of the Full Invalidation entries in entry list. The criteria for selecting the single entry is left up to the implementation to decide.

> * Otherwise if entry list contains one or more patch map entries which have a patch format that is Partial Invalidation then, select exactly one of the Partial Invalidation entries in entry list. The criteria for selecting the single entry is left up to the implementation to decide.

> * Otherwise select exactly one of the No Invalidation entries in entry list. The criteria for selecting the single entry is left up to the implementation to decide.

Why aren't we instructing people to make (up to 3) lists grouped by patch encoding, and then process those lists in the specified order? You'd never want to put them in a single list in practice and I don't see what the generality is buying us here.

skef:
If the concern is about future patch formats that might share invalidation categories, I'm happy to have the lists grouped by invalidation category rather than format.

garretrieger:
There's no particular reason for them to be merged into a single list, it should be functionally equivalent to have the text say to separate them into invalidation groups. I can update the text.

garretrieger:
Related #239

garretrieger:
Should be addressed by the #244, but feel free to reopen if more is needed.

Issue #232, opened 2024-11-08
Performance of unified extension algorithm for "no invalidation" patches


skef:
It still seems to me that the patch extension algorithm is designed around the needs of what we're now calling "table-keyed"/invalidating or partially invalidating patches, with non-invalidating patches added on. This issue (to the extent it is one) is related to my questions about the unified algorithm but not equivalent to it -- perhaps a different unified algorithm would have a different balance of focus.

In general we expect to have a smaller choice of table-keyed patches at a given level, although strategies to provide a "one round trip" solution with patches of different grains might increase the number. There could be quite a few glyph-keyed patches depending on the degree to which the encoding is trying to squeeze the transfer size to a minimum. And if you have a large file you could be loading quite a few glyph-keyed patches.

With that in mind, this is the selection description for no invalidation patches:

> Otherwise select exactly one of the [No Invalidation](https://w3c.github.io/IFT/Overview.html#no-invalidation) entries in entry list. The criteria for selecting the single entry is left up to the implementation to decide.

Later language clarifies that patch *loads* for such patches can be done concurrently. 

As stated, even if the selection is "left up to the implementation" (something we'll probably need to change) the process is specified to be in relation to the current patch subset definition, which is updated on each round. So *as specified* the update algorithm compares $\frac{(m+1)(2n-m)}{2}$ times, where $n$ is the number of patches in the list and $m$ is the number of patches that need to be loaded (fourth edit's the charm?). That's not quite $O(n^2)$ but it's in the ballpark. 

There *may* be some corner cases with glyph-keyed patches where it would make sense to pick one over another. The main case I can think of is if you have a glyph duplicated among several patches that is itself directly associated with a codepoint, and you've thrown that codepoint into the list for each patch. In such cases you might be able to avoid loading a patch by being careful about the analysis. However, those cases will be quite rare and aren't important to optimize for. So I would be much more inclined to just treat Non-invalidating patches as truly independent during selection.

garretrieger:
I'm currently working through a client side implementation of the extension algorithm so I can provide some perspective on how that looks in practice. Here's what my current implementation does:

1. locate all intersecting patches as per the procedure in the spec.
2. In one linear scan select a maximal set of patches which don't invalidate each other. That could be
- One fully invalidating patch
- One partial invalidating patch + zero or more non invalidating with differing compat ids.
- Up to two partially invalidating patches if they have different compat ids.
4. Initiate loads for all selected patches.
[...]

PR #244, opened 2024-11-22
Update extension algorithm to include explicit prefetching.

garretrieger:
- In the algorithm steps add text saying the client can optionally start loads for any patches not invalidated by the current selection.
- Group patches by invalidation category prior to the selection step to improve clarity.
- Note that intersection doesn't need to be repeated on each iteration when only no invalidation patches are left.

Addresses #232, #233, and #239.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/244.html" title="Last updated on Nov 22, 2024, 11:22 PM UTC (4f22026)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/244/65bf200...4f22026.html" title="Last updated on Nov 22, 2024, 11:22 PM UTC (4f22026)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -452,30 +452,29 @@ The algorithm:
     <var>entry list</var>.
     
 5.  For each <var>entry</var> in <var>entry list</var> invoke [$Check entry intersection$] with <var>entry</var> and
-    <var>target subset definition</var> as inputs, if it returns false remove <var>entry</var>
-    from <var>entry list</var>.
+    <var>target subset definition</var> as inputs, if it returns false remove <var>entry</var> from <var>entry list</var>. Additionally remove
+    any entries in <var>entry list</var> which have a patch URI which was loaded and applied previously during the execution of this algorithm.
 
-6.  Remove any entries in <var>entry list</var> which have a patch URI which was loaded and applied previously during the execution
-    of this algorithm.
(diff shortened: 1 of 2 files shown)


skef (approved):
Thanks, this is a good improvement

skef:
From #239:

> instructs the client to start loads for any entries which won't be invalidated by the current selection.

I agree this is a reasonable approach, and this PR accurately describes it.  But I think we may need a bit more verbiage around this in the future. The subject comes up a bit abruptly in the algorithm description and is a bit different than how we discuss other things. I had a bit of a "wait, what?" reaction, editorially. (I'm not *sure* we'll need to change it, it just struck me that way on my reading of the PR.)

garretrieger:
> From #239:
> 
> > instructs the client to start loads for any entries which won't be invalidated by the current selection.
> 
> I agree this is a reasonable approach, and this PR accurately describes it. But I think we may need a bit more verbiage around this in the future. The subject comes up a bit abruptly in the algorithm description and is a bit different than how we discuss other things. I had a bit of a "wait, what?" reaction, editorially. (I'm not _sure_ we'll need to change it, it just struck me that way on my reading of the PR.)

I updated the intro text for the extension algorith to talk a bit about this.

PR #249, opened 2025-01-17
Some clarifications around uri templates

garretrieger:
- Clarify how id 0 during substitution is handled. Previously this was a bit ambiguous as the text stated to strip all 0 bytes.
- Specify that when applying templates the substitution may fail due to a malformed template.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/249.html" title="Last updated on Jan 17, 2025, 6:52 PM UTC (b3183d6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/249/de17c8b...b3183d6.html" title="Last updated on Jan 17, 2025, 6:52 PM UTC (b3183d6)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -1851,8 +1851,8 @@ A string ID is a sequence of bytes. Several variables are defined which are used
       The input id encoded as a [[rfc4648#section-7|base32hex]] string (using the digits 0-9, A-V) with padding
       omitted.  When the id is an unsigned integer it must first be converted to a big endian 32 bit unsigned integer,
       but then all leading bytes that are equal to 0 are removed before encoding.  (For example, when the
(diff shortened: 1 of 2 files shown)


Overview.bs
```diff
@@ -1937,6 +1946,11 @@ Some example inputs and the corresponding expansions:
     <td>14,000,000</td>
     <td>//foo.bar/1Z-A</td>
   </tr>
```
skef, line 1948: Should we use some means of indicating if the key is a string or an integer? Maybe quotes around the strings?

skef (approved):
Other than the comment question, LTGM

Overview.bs
garretrieger: Good idea, I added a type column to the example table so there's no confusion.

PR #252, opened 2025-02-05
Rework the copy indices mechanism

### LINKED ISSUE #245: Add appendix B example that shows usage of a multiple subset definition patch mapping
For example to load glyphs for a UVS selector.

garretrieger:
Instead of producing multiple subset definitions, define the intersection check in terms of whether referenced child entries would intersect. This largely provides the same functionality while overall making the mechanism simpler to reason about. Additionally it provides a path to a more efficient client side implementation, which can now cache the intersection checks on prior entries to quickly compute down stream entry intersection results. The prior approach requires in some cases fully expanding out the subset definition unions for each entry to correctly compute intersection results.

Also adds an appendix B example using child indices to demonstrate how UVS specific patches can be loaded. Fixes #245.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/252.html" title="Last updated on Feb 7, 2025, 9:55 PM UTC (4d8d534)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/252/b592442...4d8d534.html" title="Last updated on Feb 7, 2025, 9:55 PM UTC (4d8d534)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -342,19 +342,18 @@ Patch Map {#patch-map-dfn}
 A <dfn dfn>patch map</dfn> is an [[open-type/otff#table-directory|open type table]] which encodes a collection of mappings from
 [=font subset definition|font subset definitions=] to URIs which host [[#font-patch-formats|patches]] that extend the
 [=incremental font=]. A [=patch map=] table encodes a list of <dfn dfn>patch map entries</dfn>, where each entry has a key and value.
-The key is a [=font subset definition=], zero or more references to other entries and the value is a URI, the [[#font-patch-formats]]
-used by the data at the URI, and the [[#font-patch-invalidations|compatibility ID]] of the patch map table. More details of the format
-of patch maps can be found in [[#font-format-extensions]].
+The key of an entry defines the coverage of the entry, which is information on what subset definitions will match it. The value specifies
+information about the patch which should be applied when the entry is matched. [=patch map entries|Patch Map Entry=] summary:
 
-[=patch map entries|Patch Map Entry=] summary:
 <table>
   <tr><th>Key</th><th>Value</th></tr>
   <tr>
     <td>
-      * [=font subset definition=]
-      * Reference to zero or more other Patch Map Entries.
-      * Prior entry match mode, which can be either conjunctive or disjunctive.
-
+      * [=font subset definition=]: defines the basic coverage.
+      * References to zero or more child [=patch map entries|Patch Map Entry=]'s:<br/>
+         these child entries are used to extend the coverage.
+      * Child entry match mode, which can be either conjunctive or disjunctive.
+      
     </td>
     <td>
       * Patch URI
@@ -364,6 +363,8 @@ of patch maps can be found in [[#font-format-extensions]].
     </td>
   </tr>
 </table>
+
+More details of the format of patch maps can be found in [[#font-format-extensions]].
 
 Explanation of Data Types {#data-types}
 ---------------------------------------
@@ -544,11 +545,11 @@ The algorithm:
 
 2. If one or more sets checked in step 1 did not intersect, then return false for <var>intersects</var>.
 
-3. If there are no entry references in <var>mapping entry</var>, then return true for <var>intersects</var>.
+3. If there are no child entries in <var>mapping entry</var>, then return true for <var>intersects</var>.
 
-4. Invoke [$Check entry intersection$] for each prior entry referenced in <var>mapping entry</var> with <var>subset definition</var> as an input. If prior
-    entry match mode is conjunctive, then return true for <var>intersects</var> if all invocations return true. Otherwise return true if at least one
-    invocation returns true.
+4. Invoke [$Check entry intersection$] for each child entry referenced in <var>mapping entry</var> with <var>subset definition</var>
+    as an input. If child entry match mode is conjunctive, then return true for <var>intersects</var> only if all invocations return true.
+    Otherwise return true only if at least one invocation returns true.
 
 <div class=example>
 
@@ -562,7 +563,7 @@ The algorithm:
       feature tags: {},
       design space: {}
     },
-    prior entries: {},
+    child entries: {},
     match mode: disjunctive,
     ```
     </td>
@@ -583,7 +584,7 @@ The algorithm:
       feature tags: {},
       design space: {}
     },
-    prior entries: {},
+    child entries: {},
     match mode: disjunctive,
     ```
     </td>
@@ -603,7 +604,7 @@ The algorithm:
       feature tags: {},
       design space: {}
     },
-    prior entries: {},
+    child entries: {},
     match mode: disjunctive,
     ```
     </td>
@@ -623,7 +624,7 @@ The algorithm:
       feature tags: {},
       design space: {}
     },
-    prior entries: {},
+    child entries: {},
     match mode: disjunctive,
     ```
     </td>
@@ -644,14 +645,14 @@ The algorithm:
       feature tags: {},
       design space: {}
     },
-    prior entries: [
+    child entries: [
       {
         subset definition {
           code points: {1, 2, 3},
           feature tags: {},
           design space: {}
         },
-        prior entries: {},
+        child entries: {},
         match mode: disjunctive,
       },
       {
@@ -660,7 +661,7 @@ The algorithm:
           feature tags: {},
           design space: {}
         },
-        prior entries: {},
+        child entries: {},
         match mode: disjunctive,
       },
     ],
@@ -684,14 +685,14 @@ The algorithm:
       feature tags: {},
       design space: {}
     },
-    prior entries: [
+    child entries: [
       {
         subset definition {
           code points: {1, 2, 3},
           feature tags: {},
           design space: {}
         },
-        prior entries: {},
+        child entries: {},
         match mode: disjunctive,
       },
       {
@@ -700,7 +701,7 @@ The algorithm:
           feature tags: {},
           design space: {}
         },
-        prior entries: {},
+        child entries: {},
         match mode: disjunctive,
       },
     ],
@@ -769,8 +770,8 @@ round trips.
 The following selection criteria minimizes round trips and must be used by the client when selecting a single partial or full invalidation patch in step 8
 of [$Extend an Incremental Font Subset$]:
 
-1.  For each candidate entry compute a total subset definition by unioning that entry's subset definition and the subset definition of any other entries
-     reachable via the graph of referenced prior entries.
+1.  For each candidate entry compute a total subset definition by unioning that entry's subset definition and the subset definition of all child entries
+     reachable via the graph of referenced child entries.
 
 2.  For each candidate entry compute the set intersection between the total subset definition and the target subset definition.
    
@@ -1437,20 +1438,20 @@ The algorithm:
 
   <tr>
     <td>uint8</td>
-    <td><dfn for="Mapping Entry">priorEntryMatchModeAndCount</dfn></td>
+    <td><dfn for="Mapping Entry">childEntryMatchModeAndCount</dfn></td>
     <td>
       This and the following field are used to add conditions to the intersection check for this entry that depend on whether prior entries intersect.
       The most significant bit is used to indicate the matching mode. If the bit is set the condition is conjunctive otherwise it is disjunctive. The
-      remaining 7 bits are interpreted as a unsigned integer and represent the number of entry indices in the priorEntryIndices list. This field is only
+      remaining 7 bits are interpreted as a unsigned integer and represent the number of entry indices in the childEntryIndices list. This field is only
       present if [=Mapping Entry/formatFlags=] bit 1 is set.
     </td>
   </tr>
   <tr>
     <td>uint24</td>
-    <td><dfn for="Mapping Entry">priorEntryIndices</dfn>[0b01111111 & priorEntryMatchModeAndCount]</td>
+    <td><dfn for="Mapping Entry">childEntryIndices</dfn>[0b01111111 & childEntryMatchModeAndCount]</td>
     <td>
       Each value is the index of an entry in [=Mapping Entries/entries=] array whose intersection will be checked to determine this entries intersection.
-      Only references entries that occurred prior to this [=Mapping Entry=] in [=Mapping Entries/entries=] and only present if
+      Only entries that occurred prior to this [=Mapping Entry=] in [=Mapping Entries/entries=] can be referenced. Only present if
       [=Mapping Entry/formatFlags=] bit 1 is set.
     </td>
   </tr>
@@ -1583,6 +1584,7 @@ The algorithm:
 
      *  If the returned value of ignored is false, then set the compatibility ID of the returned entry to
          [=Format 2 Patch Map/compatibilityId=] and add the  entry to <var>entry list</var>.
+         <!-- TODO XXXXX need to change ignored handling to work with the child entries mechanism. -->
 
 4.  Return <var>entry list</var>.
 
@@ -1641,15 +1643,15 @@ The algorithm:
 
 6.  If [=Mapping Entry/formatFlags=] bit 1 is set, then the copy indices list is present:
 
-    *  If the most significant bit of [=Mapping Entry/priorEntryMatchModeAndCount=] is set then set <var>entry</var>'s match mode to conjunctive,
+    *  If the most significant bit of [=Mapping Entry/childEntryMatchModeAndCount=] is set then set <var>entry</var>'s match mode to conjunctive,
          otherwise to disjunctive.
 
-    *  Read the prior entry indices list specified by [=Mapping Entry/priorEntryMatchModeAndCount=] and [=Mapping Entry/priorEntryIndices=] from
+    *  Read the child entry indices list specified by [=Mapping Entry/childEntryMatchModeAndCount=] and [=Mapping Entry/childEntryIndices=] from
          <var>entry bytes</var>.
 
-    *  The prior entry indices refer to previously loaded entries. 0 is the first [=Mapping Entry=] in <var>prior entry list</var>, 1 the second
-         and so on. For each value in [=Mapping Entry/priorEntryIndices=] locate the entry in <var>prior entry list</var> with a matching index.
-         Add a reference to that entry into </var>entry</var>. If a [=Mapping Entry/priorEntryIndices=] is greater than or equal to the length
+    *  The child entry indices refer to previously loaded entries. 0 is the first [=Mapping Entry=] in <var>prior entry list</var>, 1 the second
+         and so on. For each value in [=Mapping Entry/childEntryIndices=] locate the entry in <var>prior entry list</var> with a matching index.
+         Add a reference to that entry into </var>entry</var>. If a [=Mapping Entry/childEntryIndices=] is greater than or equal to the length
          of <var>prior entry list</var> then, this encoding is invalid return an error.
 
 7.  If [=Mapping Entry/formatFlags=] bit 2 is set, then an id delta or id string length is present:
@@ -2660,7 +2662,7 @@ when loading the patches for two or more segments.  There are five main strategi
      modifier code point which triggers a glyph substitution when paired with many other code points. Given the large number of alternate
      glyphs it's desirable to keep them in their own patches which are only loaded when both the modifier code point and appropriate base
      code point(s) are present. This can be achieved by using a [[#patch-map-format-2|Format 2 Patch Map]] and multiple entry matching
-     via [=Mapping Entry/priorEntryIndices=]. For the entry one subset definition should contain the modifier code point and a
+     via [=Mapping Entry/childEntryIndices=]. For the entry one subset definition should contain the modifier code point and a
      second one has the base code point(s).
 
 4.  Alternatively, the glyph can be included in more than one of the patches that correspond to those segments at the cost of duplicating
@@ -2815,6 +2817,8 @@ path: feature-registry.html
 
 <h2 id="extension-example">
 Appendix B: Extension Algorithm Example Execution</h2>
+
+<!-- TODO add an example that utilizes prior entry indices (such as UVS specific patches)  -->
 
 This appendix provides an example of how a typical IFT font would be processed by the [[#extend-font-subset]] algorithm. In this example the IFT font
 contains a mix of both [[#table-keyed]] and [[#glyph-keyed]] patches.
--- a/Overview.bs
+++ b/Overview.bs
@@ -1566,9 +1566,11 @@ The algorithm:
 2.  If the [=Format 2 Patch Map/entryIdStringData=] offset is 0 then initialize <var>last entry id</var> to 0. Otherwise initialize
      it to an empty byte string. Set <var>current byte</var> to 0, and <var>current id string byte</var> to 0.
 
-3.  Invoke [$Interpret Format 2 Patch Map Entry$], [=Format 2 Patch Map/entryCount=] times. For each invocation:
+3.  Initialize <var>entry list</var> and <var>prior entry list</var> to empty lists.
 
-     *  pass in <var>entry list</var>, the bytes from <var>patch map</var> starting from [=Format 2 Patch Map/entries=][<var>current byte</var>] to
+4.  Invoke [$Interpret Format 2 Patch Map Entry$], [=Format 2 Patch Map/entryCount=] times. For each invocation:
+
+     *  pass in <var>prior entry list</var>, the bytes from <var>patch map</var> starting from [=Format 2 Patch Map/entries=][<var>current byte</var>] to
          the end of <var>patch map</var>,
          the bytes from <var>patch map</var> starting from [=Format 2 Patch Map/entryIdStringData=][<var>current id string byte</var>]
          to the end of <var>patch map</var> if [=Format 2 Patch Map/entryIdStringData=] is non zero,
@@ -1582,11 +1584,12 @@ The algorithm:
 
      *  Add the returned consumed id string byte count to <var>current id string byte</var>.
 
+     *  Add the returned entry to <var>prior entry list</var>.
+
      *  If the returned value of ignored is false, then set the compatibility ID of the returned entry to
          [=Format 2 Patch Map/compatibilityId=] and add the  entry to <var>entry list</var>.
-         <!-- TODO XXXXX need to change ignored handling to work with the child entries mechanism. -->
 
-4.  Return <var>entry list</var>.
+5.  Return <var>entry list</var>.
 
 <dfn abstract-op>Interpret Format 2 Patch Map Entry</dfn>
 
--- a/Overview.bs
+++ b/Overview.bs
@@ -342,16 +342,18 @@ Patch Map {#patch-map-dfn}
 A <dfn dfn>patch map</dfn> is an [[open-type/otff#table-directory|open type table]] which encodes a collection of mappings from
 [=font subset definition|font subset definitions=] to URIs which host [[#font-patch-formats|patches]] that extend the
 [=incremental font=]. A [=patch map=] table encodes a list of <dfn dfn>patch map entries</dfn>, where each entry has a key and value.
-The key is one or more [=font subset definition=] and the value is a URI, the [[#font-patch-formats]] used by the data at the URI, and
-the [[#font-patch-invalidations|compatibility ID]] of the patch map table. More details of the format of patch maps can be found
-in [[#font-format-extensions]].
+The key is a [=font subset definition=], zero or more references to other entries and the value is a URI, the [[#font-patch-formats]]
+used by the data at the URI, and the [[#font-patch-invalidations|compatibility ID]] of the patch map table. More details of the format
+of patch maps can be found in [[#font-format-extensions]].
 
 [=patch map entries|Patch Map Entry=] summary:
 <table>
   <tr><th>Key</th><th>Value</th></tr>
   <tr>
     <td>
-      * One or more [=font subset definition=]
+      * [=font subset definition=]
+      * Reference to zero or more other Patch Map Entries.
+      * Prior entry match mode, which can be either conjunctive or disjunctive.
 
     </td>
     <td>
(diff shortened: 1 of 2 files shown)


svgeesus (approved):
This seems good (I read the preview, because the diff was hard for me to follow in places) and seems to be clearer than the previous method.

Overview.bs
```diff
@@ -734,17 +770,18 @@ round trips.
 The following selection criteria minimizes round trips and must be used by the client when selecting a single partial or full invalidation patch in step 8
 of [$Extend an Incremental Font Subset$]:
 
-1.  For each candidate entry compute the set intersection between each subset definition in the entry and the target subset definition. Union
-     the resulting intersections together into a single subset definition.
+1.  For each candidate entry compute a total subset definition by unioning that entry's subset definition and the subset definition of all child entries
```
skef, line 773: I'm confused about this. Why are we unioning if and when the child entry match mode is conjunctive?

Overview.bs
```diff
@@ -540,8 +543,13 @@ The algorithm:
      When checking design space sets for intersection, they intersect if there is at least one pair of intersecting segments
      (tags are equal and the ranges intersect).
 
-2. If all sets checked in step 1 intersect, then return true for <var>intersects</var> otherwise false.
+2. If one or more sets checked in step 1 did not intersect, then return false for <var>intersects</var>.
```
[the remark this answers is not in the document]
garretrieger: I intended the match mode setting to apply only to the child entries, but I can see an argument for it applying to the entry as a whole (and that would actually match the previous behaviour). I don't have a specific use case in mind for the way I've currently specified it so I'm good with changing to apply to the entry as a whole.

Overview.bs
```diff
@@ -540,8 +543,13 @@ The algorithm:
...
+
+3. If there are no child entries in <var>mapping entry</var>, then return true for <var>intersects</var>.
 
+4. Invoke [$Check entry intersection$] for each child entry referenced in <var>mapping entry</var> with <var>subset definition</var>
```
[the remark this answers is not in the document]
garretrieger: Cycles are prevented by requiring child entry indices to always point to previous entries. Specifically this requirement: "If a childEntryIndices is greater than or equal to the length of prior entry list then, this encoding is invalid return an error.". This requirement ensures the graph formed by them is a DAG.

Overview.bs
garretrieger: For this procedure we are trying to assess what code points, features, and design space an entry has data for. As an example whether a patch is loaded when (A or B), or when (A and B) both cases indicate that it has data related to both A and B.

Overview.bs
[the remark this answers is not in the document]
garretrieger: Switched to have the whole entry respect the match mode and added an example illustrating a disjunctive match with the entry's subset def and a child entry.

Overview.bs
```diff
@@ -632,18 +640,32 @@ The algorithm:
...
+      code points: {},
+      feature tags: {},
+      design space: {}
+    },
+    child entries: [
       {
-        code points: {1, 2, 3},
-        feature tags: {},
-        design space: {}
+        subset definition {
```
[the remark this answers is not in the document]
garretrieger: I changed this to treat the whole table as a series of entries in a single patch map (with assigned indices) and then specified child entries via indices as it would be done in the encoding. What do you think?

Overview.bs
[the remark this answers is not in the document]
garretrieger: Actually ran into an issue with this so going to switch back to the way it was. The problem is that it's pretty typical to have an entry that has an empty subset definition and then list its conditions using only child entries. In the disjunctive case if we treat the whole entry as disjunctive then the entry's empty subset definition always matches everything (since empty sets are considered wild cards for matching) and since it's treated disjunctively that means the entry will always match everything as well.

Overview.bs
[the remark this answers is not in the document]
skef: OK - in that case do we have potential conflicts between the patch map ordering for inclusion purposes and the ordering for tie-breaking? The tie-breaking aspect is sort of a corner case so it's not obvious there would be a problem but I'm wondering if things could get tricky and/or annoying in some circumstances.

Overview.bs
```diff
@@ -734,17 +770,18 @@ round trips.
 The following selection criteria minimizes round trips and must be used by the client when selecting a single partial or full invalidation patch in step 8
 of [$Extend an Incremental Font Subset$]:
 
-1.  For each candidate entry compute the set intersection between each subset definition in the entry and the target subset definition. Union
-     the resulting intersections together into a single subset definition.
+1.  For each candidate entry compute a total subset definition by unioning that entry's subset definition and the subset definition of all child entries
+     reachable via the graph of referenced child entries.
+
+2.  For each candidate entry compute the set intersection between the total subset definition and the target subset definition.
```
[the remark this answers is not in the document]
skef: I just meant to add the colon here too.

Overview.bs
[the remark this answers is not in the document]
skef: Is this maybe a time to compromise on matching formal logic? The current convention works well for empty sets but I worry it works poorly for others -- we've basically made it so that there's no easy way for an entry to "add onto" what's included via the child entries (unless I'm misunderstanding).

Overview.bs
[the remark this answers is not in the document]
skef: Yes, that's a good solution

Overview.bs
skef: I think I understand that point broadly but now I'm worried about the overall selection system. If I'm tracking this right in my head, this is where we pick the next patch to apply, and the way we've specified things we can have competing paths to incorporating the needed codepoints. So, for example, patches can map overlapping sets of codepoints such as a patch for A and a patch for B but also a patch for A+B.

If this is in the same space we're doing that, and we union conjunctive patches, it seems like it can make patches that do something small and specific "look larger" than patches that do something big and general. 

Or am I just confusing myself?

skef:
Another thing I'm wondering:

These changes add a child patch inclusion mechanism but eliminate the ability to specify multiple definitions for an entry. That's going to work for most cases but it means that if you don't already have a child entry with the right contents you're out of luck. 

I think for the sake of simplicity of the algorithm that's still the right choice. But it might be handy to be able to have "pure child" entries -- so entries that can be included as children but will otherwise be ignored as their own entries. Seems like that addresses the just mentioned problem and also (sort of ) opens up the possibility of support for a larger domain of boolean expressions, should those be useful somehow.

(There's no "not", of course, so that domain would not include arbitrary boolean expressions. Hmm -- should there be? You wouldn't typically want a "bare" or I guess "top" not because it would draw in codepoints that aren't supported by the font. But I can imagine cases where a not would be useful in a sub-expression.)

Overview.bs
skef: Ah, I just realized that the answer is something like "we only need to use this mechanism for the glyph-keyed patches, and ordering/tie-breaking is only relevant to the table-keyed patches". So this would just be a new mine in the field of unexplored generality we are creating.

Overview.bs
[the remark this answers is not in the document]
garretrieger: I don't believe this will typically be an issue since table keyed patch entries aren't expected to make use of child entries. However if it does come up it can be resolved by using of child entries, basically you can set up all of the conditions as ignored entries at the start of the encoding and then have the non-ignored entries in whatever order you wish following the where each one references a single ignored entry from earlier which defines the desired condition.

Overview.bs
garretrieger: So I'm actually just realizing that we may want to use conjunctive conditions for table keyed patches, and I don't believe any changes are needed in this procedure for those to work correctly:

The primary goal of this process is to minimize round trips by avoiding the selection of patches that are fully encompassed by some other larger patch which will be needed (which is what the super set stuff aims to solve). One nice thing about child entries mechanism I'm just realizing is it could be used to better specify the desired selection order for table keyed patches.  For example in a case where you have table keyed patches for A, B, A+B. We've typically thought the mappings would be A -> patch for A, B -> patch for B, and A union B -> patch for A+B. Which then necessitates an appropriate entry ordering so that if a client is selecting for only A then entry for just A will have priority over A+B.

Instead if the entry for patch A+B used the condition (A and B), then the A+B patch would only be considered for loading when code points from both A and B are present. From there for this procedure will consider the union of A and B which will cause the A+B patch to be chosen over either individual A and B patches. Note how we didn't need to depend on a specific entry ordering at any point to get the desirable outcome. This also demonstrates how using A union B in this procedure for an (A and B) gets the correct outcome.

garretrieger:
> I think for the sake of simplicity of the algorithm that's still the right choice. But it might be handy to be able to have "pure child" entries -- so entries that can be included as children but will otherwise be ignored as their own entries. Seems like that addresses the just mentioned problem and also (sort of ) opens up the possibility of support for a larger domain of boolean expressions, should those be useful somehow.

We actually have this capability currently. Format 2 entries have an ignore bit (See step 11 of https://w3c.github.io/IFT/Overview.html#interpreting-patch-map-format-2). When an ignore bit is set on an entry it won't  will be excluded from the list of entries which patch selection draws from but can still be utilized as a child entry (where it will match as expected), and yeah as you suggested this opens up the possibility of constructing complex multi level boolean conditions if desired.


> (There's no "not", of course, so that domain would not include arbitrary boolean expressions. Hmm -- should there be? You wouldn't typically want a "bare" or I guess "top" not because it would draw in codepoints that aren't supported by the font. But I can imagine cases where a not would be useful in a sub-expression.)

Interesting, I don't immediately have a case for this in mind but certainly something to consider.

Overview.bs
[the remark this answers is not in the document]
garretrieger: If needed you can achieve a fully disjunctive entry by have an empty subset definition (which matches everyting) and then specifying the remaining conditions using child entries.
 
So for example

```
entry {
  code points: {a, b, c}
  child entry {
    code points: {d, e, f}
  }
  disjunctive = true (where this was intended to apply to codepoints as well)
}
```

Can be represented as:

```
entry {
  child entry {
    code points: {a, b, c}
  }
  child entry {
    code points: {d, e, f}
  }
  disjunctive = true
}
```

Overview.bs
skef: I'll have to think more about all of this. 

However, for now I'll just say that I don't see how this doesn't hurt, let alone helps, the general ordering problem. On the glyph-keyed side A+B would typically be a small thing (like adding UVSes). You're saying that on the table-keyed side A+B would typically be a big thing (patching for both A and B). That seems *at least* confusing.

garretrieger: A table keyed patch for A+B would be used in the "one round trip" encoding scheme where you want to be able to add multiple locales at once (A, B are locales here). I'm suggesting that for these cases we can encode the condition as (A and B) instead of (A union B). (A and B) better matches the behaviour we want, since the combined patch should only ever be loaded if something from both A and B is needed.

If (A union B) is used (which is how the prototype encoder currently encodes these cases), then you need to order the entries correctly (A comes be before A+B) otherwise it's possible for a target subset definition that only contains something from A to trigger the selection of the A+B patch. (A and B) on the other hand wouldn't be a candidate match for this case so no entry ordering is required to give it less priority than the A patch.

skef: But ordering, as I remember it, was just supposed to be used for tie-breaking. Are we saying now that the last matched patch would load so we no longer need a "size" or "score" for comparison?

garretrieger: We should definitely still keep the tie breaking by ordering mechanism unchanged, it's just that in this example I've described the patch map can be set up in a way where there won't be ties that need to be broken. Since (A and B) requires both codepoints from A and B to be present the intersection with target subset definition will never be equal to that of the A or B mappings on their own and thus there wouldn't ever be a tie.

garretrieger: and more generally I believe it should be possible to implement the "one round trip" scheme with conjunctive matching that doesn't have ties.

skef: This doesn't really answer my question. What I want to know is: In this new scheme, how is a sensible "next patch" chosen *in the general case* given that some small things now "look" large. You've presented some cases in which that won't matter. Is there a general argument for why this won't ever matter, or is the encoding process supposed to avoid constructing patch maps in which it will matter, in which case what is our advice about avoiding those cases?

garretrieger: I should also add that this behaviour is not newly introduced by this PR, the prior version of the selection procedure also unioned together entries with multiple subset definitions. The updated text gives equivalent results as the old text would have. So if there's still concerns it might be better to move the discussion outside of this PR into a separate issue since the issue is not specific to this PR.

skef: This PR is the first one I'm aware of considering using these more complex patch types for table-keyed patches, so I think it's reasonable to consider these things here.

> If we have two candidate entries one activated on (A or B) and one on (A and B) these are treated by the algorithm as being of the same general size (with the tie broken by entry order) since there's no way to know what the actual nature of the two patches are. 

Yes, but what if you have (B or C) in addition to (A and B), with the sum of the codepoints referenced in patch B and those in patch C being slightly lower than those referenced in A and referenced in B? Then even if patch (A and B) does something small and specific you'll load it before patch (B or C), correct? 

(And if the answer is "yes, but you'll need to load something like (A and B) later anyway because it has different code-points, then a) I don't think that's necessarily true in all cases, it depends on how patch (B or C) patches, and b) you can still have cases of overlapping codepoints among (B or C) and (A and B) such that that won't be true.)

garretrieger: > Yes, but what if you have (B or C) in addition to (A and B), with the sum of the codepoints referenced in patch B and those in patch C being slightly lower than those referenced in A and referenced in B? Then even if patch (A and B) does something small and specific you'll load it before patch (B or C), correct?
> 
> (And if the answer is "yes, but you'll need to load something like (A and B) later anyway because it has different code-points, then a) I don't think that's necessarily true in all cases, it depends on how patch (B or C) patches, and b) you can still have cases of overlapping codepoints among (B or C) and (A and B) such that that won't be true.)

Yeah in this example both would need to be loaded since (A and B) requires codepoints from A, B to be present which means both entries will match and be loaded (assuming A, B, and C are disjoint). Even in the case where the intersection of the two conditions is identical (possible if A, B, C are not disjoint) both will likely still be loaded depending on the details of the table keyed patch setup. There's three possibilities:

1. (B or C) contains the the same data for B as (A and B). In this case application of (B or C), should update the entry for (A and B) to just (A) and have a new patch that adds the remaining A data. (A) will be loaded on the next iteration if A codepoints are present.
2. (B or C) contains no shared data with (A and B). In this case application of (B or C), should not modify the (A and B) entry other then to change to a patch applicable to the new base. (A and B) will be loaded on the next iteration if applicable.
3. (B or C) contains all of the data in (A and B). In this case application of (B or C) would completely remove the (A and B) entry. Note this would be pretty strange since it's essentially implying that (A and B) doesn't actually contain data relevant to A.

Order of selection is only important in 3 and as mentioned this would be a pretty unusual case, I wouldn't expect typical encoder setups to generate this sort of thing (generally the encoder should be declaring coverage on entries that matches what's in the patch).

The invalidation patch selection algorithms main job is to try and identify cases where one patch could do the job of two or more other patches (ie. A, B, (A and B)) and preference that one for selection thereby minimizing round trips. Otherwise the particular selection order isn't important as everything that matches will be eventually loaded.

skef: I suppose I should have used different letters. All of your 1 through 3 assume coarse-grained set operations, so that you won't see, for example,  one patch with X and another patch with Y such that X ∩ Y is not empty but X is not a strict subset of Y and Y is not a strict subset of X. 

Are we *adding* that as a restriction? 

One can easily imagine someone wanting German-focused patches and Vietnamese-focused patches and not caring about encoding the strict subset (of basic Latin) separately. Are we *requiring* those common subsets to be encoded separately to avoid partial overlaps?

skef: We can add a restriction that conjunctive map entries can only be used for glyph-keyed patches. We can add a restriction (something like) that table-keyed entries that intersect with others must be super/subsets (*if* that works). Or maybe there's a different approach.

Right now we have a proposal so that patches that do little can look larger than patches that do a lot in a scenario where round-trips are important to performance. And that this is OK seems to stem from the fact that *we*, not the specification, but *we*, don't see a case where it will be a problem.

I've griped about a lot of the unexplored space we've let in to the spec but not outright objected because it seemed plausibly separable, so that someone could just stay out of that territory unless they felt like entering into it. The conjunctive entries for table-keyed patches, with the proposed comparison mechanism, are different. The ordering is *wrong*. You seem to be saying that it's irrelevant that it's wrong because in those cases you'll always need "both things" anyway. OK, then let's codify that in the spec. If we *can't* codify it, how are we so sure that the wrong ordering won't matter?

garretrieger: > I suppose I should have used different letters. All of your 1 through 3 assume coarse-grained set operations, so that you won't see, for example, one patch with X and another patch with Y such that X ∩ Y is not empty but X is not a strict subset of Y and Y is not a strict subset of X.
> 
> Are we _adding_ that as a restriction?
> 
> One can easily imagine someone wanting German-focused patches and Vietnamese-focused patches and not caring about encoding the strict subset (of basic Latin) separately. Are we _requiring_ those common subsets to be encoded separately to avoid partial overlaps?

There's no restriction on partially overlapping things, such cases are straightforward to encode and work as expected under the selection procedures (including with conjunctive matching). In the german/vietnamese example where we don't want the common stuff as a distinct patch:

Patch Map (here Common is the set of things common to German and Vietnamese):
1. Common U German -> patch that adds Common + German Specific things
2. Common U Vietnamese -> patch that adds Common + Vietnamese Specific things

Selection scenarios:
1. Target subset intersects only Common: patch 1 and 2 are tied via the selection procedure. Tie is broken to the first listed patch which is the smaller of the two. Patch 1 is selected and applied. Entry 1 is removed from the map. Entry 2 is changed to: Vietnamese -> patch that adds Vietnamese. No further patches are selected.
2. Target subset intersects Common and Vietnamese: patch 1 and 2 match however, patch 2 has a larger intersection so it's selected and applied. Entry 2 is removed from the map. Entry 1 is changed to: German -> patch that adds German.
3. Target subset intersects Common and German: handled just like 2.
4. Target subset intersects Common, German, and Vietnamese. Patch 1 and 2 match. Depending on the specifics of the intersection with target subset either 1 or 2 will be loaded first. After that, the remaining one will still match and will be applied.

This is set up using disjunctive matching on 1 and 2. If you wanted to be able to load both German and Vietnamese in a single round trip a third mapping entry could be added (Common) OR (German AND Vietnamese). Which would only activate in case 4 and due to it's larger intersection size will take precedence over patch 1 and 2.

> We can add a restriction that conjunctive map entries can only be used for glyph-keyed patches. We can add a restriction (something like) that table-keyed entries that intersect with others must be super/subsets (_if_ that works). Or maybe there's a different approach.

I don't see a need for this, there's valid use cases for conjunctive map entries and table keyed patches and there's also no need to restrict things to subsets/supersets since partially overlapping things can be handled correctly as demonstrated above.

> 
> Right now we have a proposal so that patches that do little can look larger than patches that do a lot in a scenario where round-trips are important to performance. And that this is OK seems to stem from the fact that _we_, not the specification, but _we_, don't see a case where it will be a problem.

I don't see this case being a problem in practice, do you have a specific example in mind that results a sub optimal patch selection following the selection criteria? At a higher level I don't think it's the specifications responsibility to enforce encodings to one (or minimal) round trips. We only need to provide the tools to allow encoders to achieve that outcome (which I assert we have done). Generating a high performance encoding is entirely up to the encoder implementation. Even without composite activation conditions a poorly written encoder can produce an extremely sub optimal encoding and there's nothing in the spec to prevent that. If there's some known pitfalls with complex conditions we can certainly warn about that in encoder guidance.

If an encoder wants to use high complexity conditions in conjunction with table keyed patches and it wants to minimize round trips, then it needs to reason through how those conditions will be handled by the client and encode appropriately to get the desired outcome. If an encoder implementation doesn't want to think about this, then they can follow the procedures in the encoder guidance section to produce performant table keyed mappings which don't use composite conditions.

> 
> I've griped about a lot of the unexplored space we've let in to the spec but not outright objected because it seemed plausibly separable, so that someone could just stay out of that territory unless they felt like entering into it. The conjunctive entries for table-keyed patches, with the proposed comparison mechanism, are different. The ordering is _wrong_. You seem to be saying that it's irrelevant that it's wrong because in those cases you'll always need "both things" anyway. OK, then let's codify that in the spec. If we _can't_ codify it, how are we so sure that the wrong ordering won't matter?

Conjunctive matching + table keyed entries are definitely optional in an encoder implementation if one doesn't want to deal with them, we have a procedure described in the guidance section to generate functional and performant table keyed encodings that does not use them. That said I've provided a few examples above of how conjunctive matching can be incorporated with table keyed mappings and how it results in the correct outcomes under the current selection procedure.

I don't see the concern with the ordering that the selection criteria will produce. Specific ordering doesn't matter only the total set of patches it eventually selects. If two things are eventually selected it doesn't matter which one was picked first. Ordering is only important when there exists a patch which completely encompasses other patches, then you want to pick that one over the one of the subset patches to minimize the number of loads. If such a case doesn't then the specific ordering is irrelevant.

I'm not ruling out that there may be convoluted ways to specify a mapping that could producing a suboptimal selection, but I don't see how any reasonably constructed encoding would end up in such a situation. Reasonably here is defined as an encoding that correctly updates the mappings on patch application (removing things that have been added) and doesn't needlessly split patches up  (eg. doesn't have a case like [{abc} -> patch w/ a,  {abc} -> patch w/ bc] which could be handled by a single patch or with more compact conditions [{a} -> patch w/ a, {bc} -> patch w/ bc]) If you have a more specific example that demonstrates the selection criteria making a sub optimal overall patch selection that could be helpful to understanding your concern.

skef: > I'm not ruling out that there may be convoluted ways to specify a mapping that could producing a suboptimal selection, but I don't see how any reasonably constructed encoding would end up in such a situation.

This is presumably a tautology, on your approach, because to be a "reasonably constructed" encoding would entail not being suboptimal. "They should just have been smarter." 

This whole specification, at this point, is very convoluted. *Desirable and expected* encodings are convoluted. "Convoluted" is not a relevant epithet.

I will think through some cases but maybe you could also think through some alternatives? Like, why can't conjunctive patches have an extra field indicating, in some way comparable to codepoint count, its "size"? What would that break?

skef: We're engineering a system that we expect can be used in multiple ways. We're ensuring that that system can be used in the ways we need to use it, and don't want to close off other options. I presume that one of the reasons we don't want to close off other options, among others, is that we may find that we need to take advantage of some of those other options after all. 

We have a system for ordering patches among competing options. The goal of that system is to load patches that make more sense first, where "makes more sense" is pretty close to when the patch "does more". We're technically not adding a case with these changes where the order inverts, but the case that was there only applies to independent patches, where ordering is basically irrelevant. Now it will apply to table-keyed patches, where ordering can matter.

You're saying: "show me a case where this will screw things up in practice". *I don't know, today, everything that we may need to do in practice.* I don't think you do either. One say "we'll cross that bridge when we come to it" but I simply don't believe that's how good engineering works. What makes it more likely a system will be able to do beyond what one immediately envisions is if the features make sense together. That generally speaking each has its own area of functionality and responsibility and that one can understand how that's true. 

I don't want to discover, in six months, that we have to rewrite substantial parts of the spec in order to allow some new small thing because we added a bunch of stuff that doesn't make conceptual sense on the argument "it didn't break anything at the time".

garretrieger: > > I'm not ruling out that there may be convoluted ways to specify a mapping that could producing a suboptimal selection, but I don't see how any reasonably constructed encoding would end up in such a situation.
> 
> This is presumably a tautology, on your approach, because to be a "reasonably constructed" encoding would entail not being suboptimal. "They should just have been smarter."

The invalidating selection algorithm was designed around a couple of assumptions and will produce optimal results so long as those assumptions hold. So given that here's a take at a more formal definition of when the selection criteria will produce optimal results and when it may run into trouble:

The "coverage" of an entry is defined as the union of all subset definitions of that entry and all reachable child entries. The invalidation selection algorithm was designed around these three assumptions:

1.  When patches are applied they update the fonts mapping entries to reflect the effects of their application by removing coverage from entries (and their patches) for data that has been added by the application (ie. if something adds all data needed for A, an entry for A U B would be updated to just B).
2. Each entry's coverage is "accurate", which more specifically means:
    - An entry adds data related only to it's declared coverage, (ie. if coverage is {a, b, c} and data for d is included that fails this).
    - The coverage is as minimal as possible without violating the previous point. (ie. if coverage is {a, b, c} but data for only b and c are added this is not minimal).
3. The physical order of entries in the mapping is from smallest to largest in terms of size in bytes of the associated patch.

Note: when I say "data related to" that could also be conditional data (eg. {f} AND {i} -> patch with only fi ligature data) would satisfy these conditions. The important thing here is that when applied the ligature patch would respect assumption 1 and not remove other entries that contained the f and i glyphs.

If all of the entries for invalidating patches in an encoding satisfy these three conditions, then the selection criteria will produce a selection that minimize round trips. Otherwise if an encoding does not satisfy these three conditions then there is no gaurantee the procedure will produce an optimal result, that is may take more round trips than strictly necessary. If an encoder implementation is choosing to violate one or more of these assumption they will need to be careful and work out how whatever they're doing will behave under the selection procedure.

These three conditions are straightforward for an encoder implementation to meet and align with what we consider best practices for the table keyed encodings. Additionally they're not overly restrictive and leave room for flexibility in the encoding specifics. For example no restrictions are placed on overlaps between entries. Partial, full, no overlaps, or a mix of all of those will work correctly. Likewise these restrictions don't prevent the use of conjunctive or other composite matching conditions. A nice side effect of writing the procedure in terms of coverage is that it makes reasoning about how complex conditions will be treated very straightforward.

I think it would be a good idea to add something along these lines into the encoder guidance as it provides a clear guideline for how to produce encodings that will work optimally with the selection criteria.

> I will think through some cases but maybe you could also think through some alternatives? Like, why can't conjunctive patches have an extra field indicating, in some way comparable to codepoint count, its "size"? What would that break?

The entry ordering currently provides size information. I don't see a need for an explicitly recorded size, since the selection procedure only cares about the relative size of things. Furthermore size is actually a secondary metric for determining selections used only for tie breaking. The primary and most important metric is roughly what we previously called "stuff I need" and this depends on the interaction of both the target subset definition and what's actually in the patch. We use an intersection between the target subset definition and the entry's coverage as a proxy for this. If we wanted more information available for the selection procedure that would likely need to come in the form of per codepoint information (size or otherwise), which I don't believe would materially improve selection quality and would significantly bloat the encoded mapping sizes. There's a fundamental tradeoff here where if we want to improve selection quality we need more complete information, but that information has a cost to encode. The current selection algorithm strikes a very good balance. It produces optimal results for all cases we expect to need and it operates with a minimal amount of information (no additional information beyond the coverage data which is already present).

To justify increasing encoding sizes I'd want to see concrete use cases that meaningfully improve performance to justify the increase. Basically are there table keyed encoding schemes we'd want to produce that will produce better performance then the existing "one round trip" scheme, but won't work correctly with the current selection algorithm?

> I don't want to discover, in six months, that we have to rewrite substantial parts of the spec in order to allow some new small thing because we added a bunch of stuff that doesn't make conceptual sense on the argument "it didn't break anything at the time".

At this point we know, given the current capabilities, that we can produce a performant IFT font. We have a real working proof of concept showing that.  When we did analysis on the potential performance of IFT a couple of years back we had this idea of an "ideal implementation" which was a theoretical one that transferred exactly as much data as needed in any given extension. This ideal implementation is the theoretical lower bound for performance of an IFT font. I would assert that with current state of the art encoding schemes we've developed we are starting to approach the practical lower bound (which is higher then the theoretical one since overhead is unavoidable) and all that's left are smaller incremental improvements. As an aside, once I wrap up some remaining encoder implementation work I'd like to do some analysis and see where we are with respect to the theoretical lower bounds for a variety of archetypal fonts.

The questions around the selection procedure to me are more about whether we're leaving enough flexibility in the future for these incremental improvements. The selection procedure currently rests on a few assumptions as discussed above, which in turn adds some amount of restriction on encodings. So the question is: are those restrictions reasonable, do the give enough flexibility? If we don't think they're reasonable, then lifting those restrictions can be done but will most likely impose some additional cost in the encoding size. In my opinion the restrictions are reasonable, and leave sufficient room for additional exploration and advancement in encoding schemes.

skef: I'll read through this more thoroughly, but how do we expect criterion 3 to be met for conjunctive patches in the general case when child entries must only point to previous entries? A conjunctive A, B, C patch will typically be smaller than any of those individual patches. 

From what I understand, the answer here is "sure, the algorithm was designed around these assumptions but it turns out it does't matter for these patches so we can just set aside the assumptions." 

You point out that ordering is only used for tie-breaking, which is true in the current model, but that's the problem given that our proxy for "size" isn't accurate anymore. The union criterion works for disjunctive patches, it doesn't for conjunctive patches.

garretrieger: > I'll read through this more thoroughly, but how do we expect criterion 3 to be met for conjunctive patches in the general case when child entries must only point to previous entries?

I mentioned this in one of the other comment chains but you can use ignored entries to achieve arbitrary ordering of  the non-ignored entries even when using child entries. In this case:

entry[0]: A, ignored
entry[1]: B, ignored
entry[2]: C, ignored
entry[3]: children = {0, 1, 2}, conjunctive
entry[4]: children = {0}
...

> A conjunctive A, B, C patch will typically be smaller than any of those individual patches.

In the table keyed case this isn't typically true. A conjunctive matching condition says load this thing if A, B, and C are simultaneously present. In expected usage that's because the patch contains data for all of A, B, and C and shouldn't be loaded unless all three are needed (ie. if we only have A, B don't use this patch, find another one). Condition 2 requires that the patch which includes A, B, C in it's conditions should contain data for all three. Of course there's a corner case which I mentioned above where you could have split of some same ligature data into a separate patch, but that will still resolve correctly under the procedure since the ligature data patch won't exclude the other relevant patches from selection (I don't see a reason why you'd actually do this, but it will still work).

> 
> From what I understand, the answer here is "sure, the algorithm was designed around these assumptions but it turns out it does't matter for these patches so we can just set aside the assumptions."
> 
> You point out that ordering is only used for tie-breaking, which is true in the current model, but that's the problem given that our proxy for "size" isn't accurate anymore. The union criterion works for disjunctive patches, it doesn't for conjunctive patches.

If an encoding meets the 3 conditions then the union criterion works correctly even for conjunctive matching conditions. Whether a match is conjunctive or disjunctive doesn't communicate any meaningful information for the purpose of invalidating patch selection (it's of course meaningful when we are selecting the initial candidates via intersection checks). If a condition is (A or B) absent additional information we have no reason to believe that patch will be smaller or larger than one with condition (A and B).

There are valid use cases to use either of (A or B), (A and B) with a patch that adds all data for A + B. Like wise there's cases where either of (A or B), (A and B) could be used for a more conditional type patch that contains a small amount of supplemental data (eg. ligature glyphs for AND, shared glyphs for OR). Unless we have additional data about the contents of the patch, which means more data encoded in the mapping, then we can't do anything other than conclude (for the purposes of invalidating patch selection) that the conditions are roughly equal in cost.

skef:
Just treat @skef as off the reviewer list for this PR. You have another approval.

Issue #259, opened 2025-03-03
Use of URI templates seems problematic


annevk:
Browsers don't implement a URI parser, let alone a URI template parser. Instead they use https://url.spec.whatwg.org and https://urlpattern.spec.whatwg.org. I think it would be better to build upon those primitives.

garretrieger:
Thanks for raising this. For our usage parsing is limited just to finding and replacing the substitution expressions as defined in [rfc6570](https://datatracker.ietf.org/doc/html/rfc6570#section-3) which should be simpler then general URL parsing.
Beyond that the actual URLs are just treated as opaque strings which are passed on to fetch to resolve.

We definitely want to make sure we're reusing functionality that's already present in browsers where possible though. I reviewed the two linked specs, but it doesn't look to me like those provide template substitution functionality. Is there possibly another spec thats in use in browsers which provides template substitution functionality and would be preferable to rfc6570?

annevk:
You are correct. This would require a prose-version of https://github.com/whatwg/urlpattern/issues/73. @jeremyroman @sisidovski do you think you could add that to URLPattern so the IFT specification can build upon URLPattern instead of URI templates?

sisidovski:
I'm still catching up with the IFT. @garretrieger Could you help me understand how the URI templates are used in IFT?

svgeesus:
They are used to efficiently store the url of font patches, which can be downloaded to extend an already-downloaded font.

- [Explainer](https://github.com/w3c/IFT/blob/main/IFT-Explainer.md)
- [Specification section](https://w3c.github.io/IFT/Overview.html#uri-templates)

garretrieger:
To expand on that, the idea is this:

- Inside the font we have data (patch map) which lists a set of patches that are available and the URLs at which those patches reside.
- To minimize the cost of storing potentially hundreds of URLs we assign each patch a numeric ID which then gets substituted into a URL template to produce the final URL for each patch.
- So for example we might have an IFT font with the following mapping table:

```
URL Template: https://foo.bar/patches/{id}.patch
Patch 1: load when any of unicode code points a through m are present.
...
Patch 123: load when any of unicode code points n through z are present.
```

Then if the codepoint p is present, the client would determine that patch 123 needs to be loaded. Next, it will generate the URL to load by substituting 123 into the url template to produce: https://foo.bar/patches/FC.patch (note: the numeric id 123 gets base32 encoded to FC prior to substitution).

[...]

PR #263, opened 2025-03-18
Switch to WhatWG URL and Limit URL Templates to level 1

garretrieger:
Based on the discussion in #259:
- Stop using rfc3986 and instead use whatwg URL which supersedes rfc3986.
- Limit URL templates to level one operators, this will make it trivial for client implementations to implement template substitution and avoids needing to pull in a complete general purpose rfc6570 implementation.
- Clarify percent encoding always produces upper case letters.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/263.html" title="Last updated on Mar 20, 2025, 8:04 PM UTC (57bd8f5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/263/32f7e9c...57bd8f5.html" title="Last updated on Mar 20, 2025, 8:04 PM UTC (57bd8f5)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -1043,7 +1043,7 @@ the encoded bytes at the start of every iteration to pick up any changes made in
     <td>uint8</td>
     <td><dfn for="Format 1 Patch Map">urlTemplate</dfn>[urlTemplateLength]</td>
     <td>
-      A [[!UTF-8]] encoded string. Contains a [[#url-templates]] which is used to produce URL strings associated with each entry.
+      A [[!UTF-8]] encoded string. Contains a [[#uri-templates]] which is used to produce URL strings associated with each entry.
       Must be a valid [[!UTF-8]] sequence.
     </td>
   </tr>
@@ -1214,7 +1214,7 @@ The algorithm:
         [[open-type/cmap|cmap]] table of <var>font subset</var>. Ignore any glyph indices that are not mapped by [[open-type/cmap|cmap]].
         Multiple code points may map to the same glyph id. All code points associated with a glyph should be included.
 
-    *  Convert <var>entry index</var> into a URL string by applying [=Format 1 Patch Map/urlTemplate=] following [[#url-templates]]. If
+    *  Convert <var>entry index</var> into a URL string by applying [=Format 1 Patch Map/urlTemplate=] following [[#uri-templates]]. If
         the template expansion results in an error (see: [[rfc6570#section-3]]) then, return an error.
 
     *  If the Unicode code point set is empty then, skip this <var>entry index</var>.
@@ -1241,7 +1241,7 @@ The algorithm:
     *  If [=EntryMapRecord/firstEntryIndex|EntryMapRecord::firstEntryIndex=] is greater than
         [=EntryMapRecord/lastEntryIndex|EntryMapRecord::lastEntryIndex=] this [=EntryMapRecord=] is invalid, skip it.
 
-    *  Convert <var>mapped entry index</var> into a URL string by applying [=Format 1 Patch Map/urlTemplate=] following [[#url-templates]].
+    *  Convert <var>mapped entry index</var> into a URL string by applying [=Format 1 Patch Map/urlTemplate=] following [[#uri-templates]].
         If the template expansion results in an error (see: [[rfc6570#section-3]]) then, return an error.
 
     *  If the bit for <var>mapped entry index</var> in [=Format 1 Patch Map/appliedEntriesBitMap=] is set to 1, skip this entry.
@@ -1293,7 +1293,7 @@ The algorithm:
     *  If the bit for <var>entry index</var> in [=Format 1 Patch Map/appliedEntriesBitMap=] is set to 1, skip this
         <var>entry index</var>.
 
-    *  Convert <var>entry index</var> into a URL string by applying [=Format 1 Patch Map/urlTemplate=] following [[#url-templates]].
+    *  Convert <var>entry index</var> into a URL string by applying [=Format 1 Patch Map/urlTemplate=] following [[#uri-templates]].
         If the template expansion results in an error (see: [[rfc6570#section-3]]) then, return an error.
 
     *  If the generated URL string is equal to <var>patch URL string</var> then set the bit for <var>entry index</var> in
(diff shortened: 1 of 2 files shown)


Overview.bs
```diff
@@ -1849,10 +1841,18 @@ to give the smallest encodings for most unicode code point sets typically encoun
   ```
 </div>
 
-### URI Templates ### {#uri-templates}
+### URL Templates ### {#url-templates}
+
+URL templates [[!rfc6570]] are used to convert numeric or string IDs into URL strings where patch files are located.
```
skef, line 1846: I'd prefer leaving "URI" here and then clarify that the template must expand into a URL

Overview.bs
```diff
@@ -1849,10 +1841,18 @@ to give the smallest encodings for most unicode code point sets typically encoun
...
+A string ID is a sequence of bytes. The output of template expansion is a [[!UTF-8]] encoded string.
 
-URI templates [[!rfc6570]] are used to convert numeric or string IDs into URIs where patch files are located.
-A string ID is a sequence of bytes. Several variables are defined which are used to produce the expansion of the template:
+To limit the complexity of the template expansion implementation, URL templates in IFT are restricted to only
+[[rfc6570#section-1.2|Level 1]] substitution expressions. If during expansion unsupported operators (level 2 and above)
+or an expression which doesn't match template grammar is encountered ([[rfc6570#section-3]]), then template expansion must
```
skef, line 1851: Can we phrase this positively instead of negatively? I'd like to think that a client implementer who wants to roll their own solution never needs to look at the other levels at all, and this kind of makes it sound like they should.

Overview.bs
garretrieger: Sounds good, fixed.

Overview.bs
garretrieger: Changed this to just state that anything outside of the level 1 grammar results in an error. So for client implementation that would just entail checking that nothing other than varchar's (ALPHA / DIGIT / "_" / pct-encoded) are present between {}'s.

If we wanted we could go further and state any expressions containing variables not explicitly defined by this spec also result in errors. That would then mean anything other than {id}, {id64}, {d1}, ..., {d2} results in an error. What do you think?

garretrieger: Currently variables other then the ones we define will get substituted with empty strings according to rfc6570.

skef: > If we wanted we could go further and state any expressions containing variables not explicitly defined by this spec also result in errors. That would then mean anything other than {id}, {id64}, {d1}, ..., {d2} results in an error. What do you think?

Thinking about it, I think this is the better option for two reasons:

1. In terms of documentation, if we spell this out we can just note this falls under level 1 support, and implementers who don't care much about the spec wouldn't even have to look at it.
2. In terms of implementation, if the spec changes to add new tokens in the future and there are older clients that don't check version numbers (which is pretty common for "1.0" technologies), you'll get a clean error from the client as opposed to trying and failing to load a faulty URL. 

The potential downside that occurs to me is if you want to use an rfc6570 library but can't get it to error out on missing tokens. Is that likely? It seems like one would expect to have the *option* of not just subbing in an empty string in such a library.

Overview.bs
```diff
@@ -437,7 +428,7 @@ The inputs to this algorithm are:
 
 *  <var>font subset</var>: an [=incremental font|incremental=] [=font subset=].
 
-*  <var>initial font subset URI</var>: an [[rfc3986#section-4.3|absolute URI]] which identifies the location of the initial incremental
+*  <var>initial font URL</var>: a [[whatwg-url#concept-url|URL]] which identifies the location of the initial incremental
```
svgeesus, line 431: Good. "initial font subset" implied that there was, somewhere, also an initial font which was not subsetted; this is clearer

Overview.bs
```diff
@@ -691,23 +682,24 @@ when a set (ie. code points, feature tags, design space, child entries) is not s
...
-1.  Perform [[rfc3986#section-5|reference resolution]] on <var>Patch URI</var> using <var>Initial Font URI</var> as the base URI to
-     produce the <var>target URI</var>.
+1.  Parse <var>patch URL string</var> into a [[whatwg-url#concept-url|URL Record]] using [[whatwg-url#url-parsing]].
+     <var>initial font URL</var> is the [[whatwg-url#concept-base-url|base URL]] and encoding is UTF-8. Store the result
+     in <var>target URL</var>.
 
-2.  Retrieve the contents of <var>target URI</var> using the fetching capabilities of the implementing user agent. For web browsers,
-     [[fetch]] should be used. When using [[fetch]] a request for patches should use the same CORS settings as the initial request for
+2.  Retrieve the contents of <var>target URL</var> using the fetching capabilities of the implementing user agent. For web browsers,
+     [[FETCH]] should be used. When using [[FETCH]] a request for patches should use the same CORS settings as the initial request for
```
svgeesus, line 702: why does this PR remove "fetch" from the bibliography? Is it because the specref one is more recent?

svgeesus (approved):

Overview.bs
garretrieger: > > If we wanted we could go further and state any expressions containing variables not explicitly defined by this spec also result in errors. That would then mean anything other than {id}, {id64}, {d1}, ..., {d2} results in an error. What do you think?
> 
> Thinking about it, I think this is the better option for two reasons:
> 
> 1. In terms of documentation, if we spell this out we can just note this falls under level 1 support, and implementers who don't care much about the spec wouldn't even have to look at it.
> 2. In terms of implementation, if the spec changes to add new tokens in the future and there are older clients that don't check version numbers (which is pretty common for "1.0" technologies), you'll get a clean error from the client as opposed to trying and failing to load a faulty URL.

Yeah, agreed this will definitely be better for future extensions.

> 
> The potential downside that occurs to me is if you want to use an rfc6570 library but can't get it to error out on missing tokens. Is that likely? It seems like one would expect to have the _option_ of not just subbing in an empty string in such a library.

I haven't looked at too many of the libraries but I'm guessing most don't support erroring on undefined variable names. However, the hope is with restricting this to level one and only supporting a small set of fixed variables that implementing a custom expander should be simple and straightforward. Alternatively if using a general purpose library it should be easy to scan the template for variable names that don't match the allowed set before running expansion.

I'll update this to disallow non-defined variable names.

Overview.bs
garretrieger: Yeah, I prefer to use the specref's when they are available since they stay up to date unlike our hardcoded ones.

Overview.bs
garretrieger: Ok, made the changes.

skef (approved):
LGTM

Issue #260, opened 2025-03-04
Clarify handling of CFF and CFF2


skef:
The current IFT specification barely mentions CFF or CFF2. The IFTB specification and prototype had a number of CFF-related specifics, including restrictions on the table layout (CharStrings index comes last) and an offset to the CharStrings index in the header.

I can imagine getting rid of *some* of these specifics by relegating them to the implementation. I suppose that clients could come with enough of a CFF parser to find the index, although that's actually a bit of a pain. But do we really want clients to be rewriting the whole table to deal with a CharStrings index that falls at the start or in the middle? That could be quite complicated in the arbitrary case.

It seems like the decision not to keep CFF/CFF2 in the prototype may be leaking back into the completeness of the specification.

garretrieger:
Yeah agreed we might end up needing to say more. I think a good next step would be to take a pass at adding CFF/CFF2 support into the encoder and see what issues are encountered. That should be helpful in informing any spec changes that might be needed.

skef:
> I think a good next step would be to take a pass at adding CFF/CFF2 support into the encoder and see what issues are encountered.

I would think the place to start would be the runtime rather than the encoder, unless you're thinking that we'd need source files to check the runtime.

As I see it the two main questions are:

1. Will the specification require that the CharStrings index come last in the CFF file (I'm thinking probably yes)
2. Will the IFT/IFTX table store an offset to the CharStrings index in the CFF or will clients need enough of a CFF parser to find the offset of the index. 

One could argue that the code to find the index will be somewhere in the implementation anyway but that doesn't seem relevant in practice: we either need to store the offset or include some sort of parser in the client code.

garretrieger:
Right, makes sense. Once I wrap up getting the wasm demo working again. I'll spend some time taking a pass at adding CFF patching support into the fontations client impl.

skef:
Another part of this will be loosening the offset size requirement compared with the IFTB prototype. That draft spec made it easier on the client by always requiring the large offset type for both glyf/loca and CFF. The current spec got rid of that for glyf, so that if the needed offsets cross over the 16 bit boundary the client is responsible for increasing it. I'm assuming we'll want to do the same for CFF/CFF2 for the same reasons.

[...]

PR #265, opened 2025-03-29
For fonts with CFF or CFF2 tables add a field to the patch map with o

garretrieger:
- Specified to only be on the 'IFT ' table so it's not duplicated in fonts which also have an IFTX patchmap.
- Required to be present and accurate when font has CFF and/or CFF2.

For #260

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/265.html" title="Last updated on Mar 31, 2025, 6:14 PM UTC (0ac20ee)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/265/af20875...0ac20ee.html" title="Last updated on Mar 31, 2025, 6:14 PM UTC (0ac20ee)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -951,6 +951,9 @@ simplified. They allow glyph data to be inserted into CFF/CFF2 tables without ch
 encoded in the CFF/CFF2 table. The stored CFF/CFF2 offsets allow the client to locate the CharStrings INDEX without needing to
 parse any other parts of the CFF/CFF2 table.
 
+Note: that in some cases a CFF or CFF2 table may be added or changed by a table keyed patch. In these cases the patch
+will need to also add or update the charstrings offset to reflect the new or changed contents.
+
 CFF or CFF2 subroutinization is compatible with the incremental font transfer. However, naive subroutinization will tend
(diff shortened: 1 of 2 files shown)


skef (approved):
This looks fine to me. These seem to be the only entries (at least for format 1, I didn't read back through format 2) that are conditional. Are there any issues with that? Ambiguities with dessicated fonts maybe?

garretrieger:
> This looks fine to me. These seem to be the only entries (at least for format 1, I didn't read back through format 2) that are conditional. Are there any issues with that? Ambiguities with dessicated fonts maybe?

I intentionally placed them at the end of the "header" sections of both formats so that if an encoder accidentally includes them when they're not actually needed and the client isn't looking for them they will naturally be skipped over when following the offsets for the remaining data in the table.

Good point about desiccated fonts, I think we probably need to say a bit more about the presence of these. Basically they may not be found in the initial font but if an CFF table is added later on then the char strings offsets should also be inserted into the 'IFT ' mapping table by the patch that added the CFF table.

skef:
Note looks good

PR #273, opened 2025-05-07
In format 2 interpretation handling of an unspecified entry id delta/length was missing.

garretrieger:
Add an additional step to generate the URL for that case. This makes the interpretation algorithm match the field descriptions.


skef (approved):
LGTM

PR #280, opened 2025-05-30
[IFT] add more specific information on configuring the fetch request.

### LINKED ISSUE #276: Load patch file fetching algorithms is lacking details
https://w3c.github.io/IFT/Overview.html#abstract-opdef-load-patch-file needs to describe in much more detail how the fetch is setup and what kind of parameters are passed in. Otherwise it's unclear what the destination is, whether CORS is enforced, how Referer gets set, etc.

This is a pretty important aspect of the security story of this standard, it can't be left completely undefined.

garretrieger:
Fixes #276.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/280.html" title="Last updated on Jun 5, 2025, 7:33 PM UTC (6706928)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/280/a2d65fe...6706928.html" title="Last updated on Jun 5, 2025, 7:33 PM UTC (6706928)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -728,12 +728,19 @@ The algorithm:
      *  Set [=request/referrer=] to <var>initial font URL</var>.
      *  Set [=request/mode=] to "cors".
 
-    For non [[FETCH]] implementations, make a request for <var>target URL</var> using the fetching capabilities that are available.
-    Copy the [[FETCH]] based settings specified above where equivalent concepts exist.
-   
-3.  Return the retrieved contents as <var>patch file</var>, or an error if the fetch resulted in an error.
 
-Note: the fetch settings aim to match those used for CSS font loading [[css-fonts-4#font-fetching-requirements]].
+    [=fetch|Fetch=] the [=/request=] where [=processResponseConsumeBody=] is step 3 of this algorithm.
+
+    For non [[FETCH]] implementations, make a request for <var>target URL</var> using the fetching capabilities that are available.
+    Copy the [[FETCH]] based settings specified above where equivalent concepts exist. Proceed to step 3 with the results of that
+    fetch.
+   
+3.  [=processResponseConsumeBody=] is passed the response and null, failure, or a byte sequence. If a byte sequence is received,
+    then return the bytes as <var>patch file</var>. If null is received, then return an empty byte array as <var>patch file</var>.
+    Otherwise return an error.
+
+Note: the fetch settings aim to match those used for CSS font loading [[css-fonts-4#font-fetching-requirements]], but with a couple of differences:
+first referrer is set to the initial font's URL, second URL resolution uses the initial font as the base instead of the stylesheet.
 
 <dfn abstract-op>Handle errors</dfn>
 
--- a/Overview.bs
+++ b/Overview.bs
@@ -701,7 +701,7 @@ when a set (ie. code points, feature tags, design space, child entries) is not s
 
 The inputs to this algorithm are:
 
-*   <var>patch URL string</var>: A [[whatwg-url|URL]] string ([[!UTF-8]] encoded) identifying the patch file to load. It  may be a
+*  <var>patch URL string</var>: A [[whatwg-url|URL]] string ([[!UTF-8]] encoded) identifying the patch file to load. It  may be a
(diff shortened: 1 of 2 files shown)


skef (approved):
I'm not an expert on this but editorially this looks good to me.

svgeesus (approved):
This looks good, and consistent with what CSS uses, but I defer to @noamr for final check

noamr:
I think it would be better to reference https://drafts.csswg.org/css-fonts-4/#fetch-a-font instead of performing your own fetch, or https://drafts.csswg.org/css-values-4/#fetch-a-style-resource if this doesn't work exactly like a font?
If there is a reason not to use either, it should at least look like them... I'm missing a line like [this](https://drafts.csswg.org/css-values-4/#fetch-a-style-resource:~:text=Fetch%20req%2C%20with%20processResponseConsumeBody%20set%20to%20processResponse) for example.

garretrieger:
> I think it would be better to reference https://drafts.csswg.org/css-fonts-4/#fetch-a-font instead of performing your own fetch, or https://drafts.csswg.org/css-values-4/#fetch-a-style-resource if this doesn't work exactly like a font? If there is a reason not to use either, it should at least look like them... I'm missing a line like [this](https://drafts.csswg.org/css-values-4/#fetch-a-style-resource:~:text=Fetch%20req%2C%20with%20processResponseConsumeBody%20set%20to%20processResponse) for example.

That's a good idea. The intention is for this to match the font fetch (with the exception of referrer being set to the parent font instead of the style sheet), I'll take a pass at rewriting this to directly call either the fetch a font or fetch a style resource algorithms.

garretrieger:
> > I think it would be better to reference https://drafts.csswg.org/css-fonts-4/#fetch-a-font instead of performing your own fetch, or https://drafts.csswg.org/css-values-4/#fetch-a-style-resource if this doesn't work exactly like a font? If there is a reason not to use either, it should at least look like them... I'm missing a line like [this](https://drafts.csswg.org/css-values-4/#fetch-a-style-resource:~:text=Fetch%20req%2C%20with%20processResponseConsumeBody%20set%20to%20processResponse) for example.
> 
> That's a good idea. The intention is for this to match the font fetch (with the exception of referrer being set to the parent font instead of the style sheet), I'll take a pass at rewriting this to directly call either the fetch a font or fetch a style resource algorithms.

In looking into the https://drafts.csswg.org/css-values-4/#fetch-a-style-resource a bit more, I think it might not be a good fit for calling from here:
- We want to set the referer differently (to the font instead of the style sheet).
- We want to resolve the url differently (with the font as the base instead of the style sheet).

So I think it's probably best to stick with the current approach, but I'll update it to include referencing process response.

garretrieger:
> > > > I think it would be better to reference https://drafts.csswg.org/css-fonts-4/#fetch-a-font instead of performing your own fetch, or https://drafts.csswg.org/css-values-4/#fetch-a-style-resource if this doesn't work exactly like a font? If there is a reason not to use either, it should at least look like them... I'm missing a line like [this](https://drafts.csswg.org/css-values-4/#fetch-a-style-resource:~:text=Fetch%20req%2C%20with%20processResponseConsumeBody%20set%20to%20processResponse) for example.
> > > 
> > > 
> > > That's a good idea. The intention is for this to match the font fetch (with the exception of referrer being set to the parent font instead of the style sheet), I'll take a pass at rewriting this to directly call either the fetch a font or fetch a style resource algorithms.
> > 
> > 
> > In looking into the https://drafts.csswg.org/css-values-4/#fetch-a-style-resource a bit more, I think it might not be a good fit for calling from here:
> > 
> > * We want to set the referer differently (to the font instead of the style sheet).
> > * We want to resolve the url differently (with the font as the base instead of the style sheet).
> > 
> > So I think it's probably best to stick with the current approach, but I'll update it to include referencing process response.
> 
> Ok cool Also you need to explicitly call [`[=fetch=]`](https://fetch.spec.whatwg.org/#concept-fetch), referencing the `[[FETCH]]` spec is not enough.

Sounds good, just pushed an update to this PR to do that. Also added a note in the text that highlights how this differs from the css font fetch.

noamr:
Much better!

PR #293, opened 2025-08-11
Create conform.css

svgeesus:
Copy of https://github.com/w3c/woff/blob/main/woff2/conform.css


svgeesus:
Actually I see the recent changes don't include the `ua`, `ff` `at` and `dc` classes to converting to draft while we discuss what styling we want.

svgeesus:
User Agent, File Format, Authoring Tool, and erm I forget dc sorry

PR #301, opened 2025-09-03
Add client/encoder tags to the conformance spans.

garretrieger:
Also move conform styles into the spec document since conform.css isn't being included.

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -82,8 +82,10 @@ spec: i18n-glossary; urlPrefix: https://www.w3.org/TR/i18n-glossary/; type: dfn;
 </pre>
 
 <style>
-.conform:hover {background: #31668f; color: white}
-.conform:target {padding: 2px; border: 2px solid #AAA; background: #31668f; color: white }
+.conform:hover {background: #DFD}
+.conform:target {padding: 2px; border: 2px solid #ACA; background: #DFD }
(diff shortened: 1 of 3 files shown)


svgeesus (commented):
I started marking conformance statements that were also an encoder requirement. But there were a lot, so I stopped to bring it back to discussion.. Also, unless the spec says what to do when each of these is not met, it isn't really a client requirement.

garretrieger:
> I started marking conformance statements that were also an encoder requirement. But there were a lot, so I stopped to bring it back to discussion.. Also, unless the spec says what to do when each of these is not met, it isn't really a client requirement.

So for the set of requirements that appear in the format 1 and 2 descriptions I believe we should tag these as both client and encoder requirements.

On the encoder side they are referenced from point 1 in the encoder requirement section.

On the client side, they are tied to client side behaviour by the interpretation algorithm which says that the client should error out if any of the format requirements are not satisfied. Example wording:

"Check that the patch map data is: complete and not truncated, has [format](https://w3c.github.io/IFT/Overview.html#format-1-patch-map-format) equal to 1, and is valid according to the requirements in [§ 5.3.1 Patch Map Table: Format 1](https://w3c.github.io/IFT/Overview.html#patch-map-format-1) (requirements are marked with a "must"). If it is not return an error."

PR #302, opened 2025-10-03
Add conform, client classes to normative algorithms.

garretrieger:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/pull/302.html" title="Last updated on Oct 6, 2025, 7:04 PM UTC (3acd670)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IFT/302/b871826...3acd670.html" title="Last updated on Oct 6, 2025, 7:04 PM UTC (3acd670)">Diff</a>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -421,7 +421,7 @@ should typically include all features found in [[#feature-tag-list]] in the subs
 know that the specific shaper which will be used may not make use of some features in [[#feature-tag-list]] and may
 opt to exclude those unused features from the subset definition.
 
-<h3 algorithm id="extend-font-subset">Incremental Font Extension Algorithm</h3>
+<h3 algorithm id="extend-font-subset" class="conform client">Incremental Font Extension Algorithm</h3>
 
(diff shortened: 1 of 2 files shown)


skef (commented):
The changes to the markdown look fine to me. There are some seemingly unrelated changes to the HTML that probably have to do with bikeshed versions, but what's up with the seemingly contradictory stuff about implementation reports on 829 and 852?

svgeesus (commented):
So this change seems to be preventing conformance styling beingf applied to headings (because it changes the selectors to only apply to spans). Is that the intent?

garretrieger:
> The changes to the markdown look fine to me. There are some seemingly unrelated changes to the HTML that probably have to do with bikeshed versions, but what's up with the seemingly contradictory stuff about implementation reports on 829 and 852?

Yeah it's being applied automatically by bikeshed, I've been fixing it up manually in the last couple of PRs and forgot to do it for this one. I believe there's a template we need to update. I've been meaning to track that down and get it fixed. For now I corrected the HTML manually.

skef (approved):
LGTM

PR #310, opened 2026-06-01
Update the default feature registry to include palt.

terkel:
Add `palt` (Proportional Alternate Widths) to the default feature registry by flipping its `Is Default` flag from `0` to `1`.

`palt` is now included in HarfBuzz's default subset layout features ([harfbuzz#6007](https://github.com/harfbuzz/harfbuzz/pull/6007), resolving [harfbuzz#6006](https://github.com/harfbuzz/harfbuzz/issues/6006)), and `feature-registry.csv` is derived from `hb-subset-input.cc`. This keeps the registry in sync with its source of truth, mirroring #306 which added `halt`/`vhal`. `palt` is one of the most commonly used optional features in Japanese typography, so subsetting silently dropping it is a real problem in practice.

I haven't regenerated `Overview.html` — it looks like the HTML rebuilds are usually done separately by maintainers (e.g. #309, #292). Happy to regenerate it in this PR if you'd prefer.

### DIFF
--- a/Overview.html
+++ b/Overview.html
@@ -5,9 +5,9 @@
   <title>Incremental Font Transfer</title>
   <meta content="ED" name="w3c-status">
   <link href="https://www.w3.org/StyleSheets/TR/2021/W3C-ED" rel="stylesheet">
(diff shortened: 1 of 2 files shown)


garretrieger (approved):
html changes from bikeshed update look good to me.

PR #311, opened 2026-07-27
Remove Patch Map Format 1 from the specification.

garretrieger:
Now that the encoder is much more developed, we've found that encodings rarely are eligible to use format 1. As a result it's unlikely format 1 would be used in practice. So drop format 1 from the specification with format 2 becoming the sole patch map format.

Format 2 is a superset of format 1's functionality, so this does not cause any loss of functionality.

Discussed removal on the mailing list: https://lists.w3.org/Archives/Public/public-webfonts-wg/2026Jul/0011.html

***

### :boom: Error: 422 Unprocessable Entity :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jul 31, 2026, 11:52 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/publications/spec-generator/) - Spec Generator is the web service used to build bikeshed/ReSpec specs

:link: [Related URL](https://www.w3.org/publications/spec-generator/?type=bikeshed-spec&output=html&url=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2FIFT%2Fe8f8a36eb983ea2f3a0c0918a20fd03446134191%2FOverview.bs&force=1&md-warning=not%20ready)

**Error output:**

```json
[
    {
        "lineNum": "2824:1",
        "messageType": "fatal",
        "text": "Couldn't find include file 'feature-registry.html'."
    },
    {
        "lineNum": null,
        "messageType": "failure",
        "text": "Did not generate, due to errors exceeding the allowed error level."
    }
]
```

_This seems to be an issue with the [Spec Generator](https://www.w3.org/publications/spec-generator/) service. PR Preview doesn't manage this service and so has no control over it. If you've identified an issue with it, you can [report the issue to the maintainers of Spec Generator](https://github.com/w3c/spec-generator/issues/new) directly. Please be courteous. Thank you!_

_If you don't have enough information above to solve the error by yourself or if the issue doesn't seem related to Spec Generator, you can [file an issue with PR Preview](https://github.com/tobie/pr-preview/issues/new?title=Unidentified%20Error&body=See%20w3c/IFT%23311.)._

</details>

### DIFF
--- a/Overview.bs
+++ b/Overview.bs
@@ -1016,7 +1016,7 @@ bytes at the start of every iteration to pick up any changes made in the previou
   <tr>
     <td>uint8</td>
     <td><dfn for="Patch Map Table">format</dfn></td>
-    <td>Set to 2, identifies the format of this table.</td>
+    <td>Set to 2, identifies the format of this table. Note: prior versions of the spec had a format 1 which has been removed, to maintain compatibility and provide a future extension point the format field has been retained with it's previous value of 2.</td>
   </tr>
   <tr>
     <td>uint24</td>
--- a/Overview.bs
+++ b/Overview.bs
@@ -460,7 +460,7 @@ The algorithm:
 3.  If the compatibility ID in 'IFT ' is equal to the compatibility ID in 'IFTX' this is an error, invoke [$Handle errors$].
 
 4.  For each of [[open-type/otff#table-directory|tables]] 'IFT ' and 'IFTX' (if present): convert the table into a list of entries by
-    invoking [$Interpret Format 1 Patch Map$] or [$Interpret Format 2 Patch Map$]. Concatenate the returned entry lists into a single list,
+    invoking [$Interpret Patch Map$]. Concatenate the returned entry lists into a single list,
     <var>entry list</var>.
     
 5.  For each <var>entry</var> in <var>entry list</var> invoke [$Check entry intersection$] with <var>entry</var> and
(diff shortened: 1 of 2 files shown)


Overview.bs
```diff
@@ -1645,39 +1274,39 @@ being requested.
...
 
 The algorithm outputs:
 
 * <var>entry list</var>: a list of [=patch map entries=].
 
 The algorithm:
 
-1.  Check that the <var>patch map</var> has [=Format 2 Patch Map/format=] equal to 2 and is valid according to the requirements in
-     [[#patch-map-format-2]] (requirements are marked with a "must"). If it is not return an error.
+1.  Check that the <var>patch map</var> has [=Patch Map Table/format=] equal to 2 and is valid according to the requirements in
```
svgeesus, line 1293: Is there value in having a `format=2`?

svgeesus (commented):
Looks good in general. Is the `format=2` an extensibility point in case we add another patch map format later?

Overview.bs
svgeesus: Discussed on the [30 July 2026 call](https://www.w3.org/2026/07/30-webfonts-minutes.html#c80e)

skef (approved):
LGTM

